PR #9, opened 2019-07-17
Add CopyIcon component and update ClipboardButton

siegfried-ehret-sonarsource:
This allows to fix https://jira.sonarsource.com/browse/SONAR-12199

Screenshot 2019-07-17 at 16 28 07

TODO:

- [ ] create a new version of the lib


components/icons/CopyIcon.tsx
```diff
@@ -0,0 +1,35 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import * as React from 'react';
+import Icon, { IconProps } from './Icon';
+
+export default function CopyIcon({ className, fill = '#777', size = 16 }: IconProps) {
```
gregaubert: I'm wondering about the default fill color here... I think it should be `'currentColor'` instead. And the specific color should be passed per usage if needed.

If we really want this specific color by default, the theme should probably be used, because it's the `secondFontColor`.

components/icons/CopyIcon.tsx
```diff
@@ -0,0 +1,35 @@
...
+  return (
+    <Icon className={className} size={size} viewBox="0 0 32 32">
```
gregaubert: Why overriding the `viewBox` ? Apart from a few exceptions all our icons have a `viewBox` of 16 per 16.
You should resize the svg to match this `viewBox`.

components/icons/CopyIcon.tsx
```diff
@@ -0,0 +1,35 @@
...
+      <g>
```
gregaubert: To verify but I think the group is useless here

components/controls/ClipboardButton.tsx
```diff
@@ -65,25 +66,41 @@ export default class ClipboardButton extends React.PureComponent<Props, State> {
...
+        onMouseEnter={this.handleMouseEnter}
+        onMouseLeave={this.handleMouseLeave}
+        overlay={translate(this.state.isCopied ? 'copied_action' : 'copy_to_clipboard')}
+        visible={this.state.tooltipShown}>
         <Button
           className={classNames('no-select', this.props.className)}
           data-clipboard-text={this.props.copyValue}
           innerRef={node => (this.copyButton = node)}>
-          {this.props.label ? this.props.label : translate('copy')}
+          {this.props.children || (this.props.label ? this.props.label : translate('copy'))}
```
gregaubert: Let's completely drop label and always use children instead. And if there are no children prop we display the `translate('copy')`.

components/controls/ClipboardButton.tsx
```diff
@@ -65,25 +66,41 @@ export default class ClipboardButton extends React.PureComponent<Props, State> {
...
       }, 1000);
     }
   };
 
   render() {
     return (
-      <Tooltip overlay={translate('copied_action')} visible={this.state.tooltipShown}>
+      <Tooltip
+        onMouseEnter={this.handleMouseEnter}
+        onMouseLeave={this.handleMouseLeave}
```
gregaubert: I'm really skeptical about exposing those internal stuff... I think we should never have to care about that when using the tooltip.

I think it would be way simpler to play with the visible state, switching it between `true` to force it visible when copied and `undefined` to have the default mouseover behavior.

Don't hesitate to ping me if you want to discuss further the implem.

components/controls/Tooltip.tsx
```diff
@@ -30,6 +30,8 @@ interface Props {
   children: React.ReactElement<{}>;
   mouseEnterDelay?: number;
   mouseLeaveDelay?: number;
+  onMouseEnter?: VoidFunction;
+  onMouseLeave?: VoidFunction;
```
gregaubert: Like I said in my other comment, let's not expose this if possible.

gregaubert (changes requested):

components/controls/ClipboardButton.tsx
wouter-admiraal-sonarsource: Let's not forget what the original intent was: to show a confirmation message when clicked (after click, it displays a tooltip with "Copied to Clipboard" for 1s). Now, we're adding a tooltip to explain what it does as well...

Probably that line 97 is sufficient:

```ts
overlay={translate(this.state.isCopied ? 'copied_action' : 'copy_to_clipboard')}
```

That way, we leave the enter/leave logic in Tooltip (as it should, completely agree with Greg), and the message is still displayed correctly once clicked.

wouter-admiraal-sonarsource (approved):
LGTM!

components/icons/CopyIcon.tsx
```diff
@@ -0,0 +1,34 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import * as React from 'react';
+import Icon, { IconProps } from './Icon';
+
+export default function CopyIcon({ className, fill = 'currentColor', size = 16 }: IconProps) {
```
gregaubert: 16 is already the default value.

components/icons/CopyIcon.tsx
```diff
@@ -0,0 +1,34 @@
...
+  return (
+    <Icon className={className} size={size}>
+      <path
+        d="M630,248h-8a.945.945,0,0,0-1,1v9h1v-9h8Zm-6,2a.945.945,0,0,0-1,1v10c0,.733.267,1,1,1h8c.733,0,1-.267,1-1V251c0-.733-.267-1-1-1Zm0,11h8V251h-8Z"
+        fillRule="evenodd"
+        style={{ fill }}
+        transform="translate(-619 -247)"
```
gregaubert: transformations can be cleaned (applied to the d property) by using something like [svgomg](https://jakearchibald.github.io/svgomg/)

components/controls/ClipboardButton.tsx
```diff
@@ -67,23 +66,23 @@ export default class ClipboardButton extends React.PureComponent<Props, State> {
 
   showTooltip = () => {
     if (this.mounted) {
-      this.setState({ tooltipShown: true });
+      this.setState({ isCopied: true });
       setTimeout(() => {
         if (this.mounted) {
-          this.setState({ tooltipShown: false });
+          this.setState({ isCopied: false });
```
gregaubert: I'm not sure it's really working as expected... So now the current workflow is:
1. mouse over we display: "Click to copy to clipboard"
2. when I click we display: "Copied to clipboard"
3. after 1 second the message change back to: "Click to copy to clipboard"

I'm really not sur about that last part, I think this timer was used to display and hide after a little time, now it's used to change the text... Should we really change the text based on a timer ? 

Also I think we should keep the flag to force show the tooltip for a second (better ux when using a keyboard, the tooltip will be displayed).

gregaubert (changes requested):
I'm not sure about the current way those tooltips work now, maybe we should discuss with @mathieu-cutivel-sonarsource  ?

components/icons/CopyIcon.tsx
siegfried-ehret-sonarsource: Already done :)

components/controls/ClipboardButton.tsx
wouter-admiraal-sonarsource: Mixing displaying on hover and with a timer is going to introduce a new layer of complexity. I personally don't mind having the tooltip remaining visible. The user will probably move away her mouse anyway.

gregaubert:
Like I was saying offline, I wonder if we should have two components for that. One for the icon copy button and one for normal copy button.

Because the tooltips that was displayed for 1 second after the click was basically there to give a visual feedback once you click on the normal button, otherwise nothing happens... And we I don't recall having permanent tooltips anywhere on normal buttons, their label should be explicit enough (if not we usually add a "?" icon with the help tooltip next to it)...

But the icon copy button is a different beast and obviously needs a tooltip to know what it's gonna do, and indeed I don't think it should have a timer that adds complexity...

WDYT @siegfried-ehret-sonarsource @wouter-admiraal-sonarsource  ?

wouter-admiraal-sonarsource:
> Like I was saying offline, I wonder if we should have two components for that. One for the icon copy button and one for normal copy button.
> 
> Because the tooltips that was displayed for 1 second after the click was basically there to give a visual feedback once you click on the normal button, otherwise nothing happens... And we I don't recall having permanent tooltips anywhere on normal buttons, their label should be explicit enough (if not we usually add a "?" icon with the help tooltip next to it)...
> 
> But the icon copy button is a different beast and obviously needs a tooltip to know what it's gonna do, and indeed I don't think it should have a timer that adds complexity...
> 
> WDYT @siegfried-ehret-sonarsource @wouter-admiraal-sonarsource ?

And what if we add a `title` attribute to this one? I know it's not a pattern we have in other places... But it would more or less solve our problem.

gregaubert:
> And what if we add a title attribute to this one? I know it's not a pattern we have in other places... But it would more or less solve our problem.

Mmmh I'm not a fan, that would mean having the title on both a button and icon button, and they would be the only one of the whole app...

siegfried-ehret-sonarsource:
@gregaubert @wouter-admiraal-sonarsource do you have more things to review on this ?

wouter-admiraal-sonarsource:
Have you discussed this with Matt? Apparently, the current solution is fine UX-wise. Anyway, my approval remains: LGTM 👍

wouter-admiraal-sonarsource (approved):

siegfried-ehret-sonarsource:
Yep, it has been discussed with Matt.

gregaubert:
I wasn't sur if it was re-ready to review, I'll do it now. I had a short discussion with Matt yesterday too, and he also said that current behavior of the "normal copy button" should not change because of the new "icon copy button". So I'll check if it's the case.

components/icons/CopyIcon.tsx
gregaubert: Seems like it's not done yet ? on the current head of your PR ? I still see the "transform", normally svgomg should take care of removing it. Unless it's not able to ?

components/controls/ClipboardButton.tsx
```diff
@@ -19,52 +19,20 @@
...
-  label?: string;
-}
-
 interface State {
   tooltipShown: boolean;
 }
 
-export default class ClipboardButton extends React.PureComponent<Props, State> {
-  private clipboard?: Clipboard;
+export default class ClipboardButton extends ClipboardBase {
```
gregaubert: Oh ? This does actually work !? First time I actually see inheritance with components...
I don't know if it's a good idea or not, here is Facebook takes on it: https://reactjs.org/docs/composition-vs-inheritance.html#so-what-about-inheritance

components/controls/ClipboardIconButton.tsx
```diff
@@ -0,0 +1,64 @@
...
+  render() {
+    const tooltipProps = {
+      overlay: translate(this.state.isCopied ? 'copied_action' : 'copy_to_clipboard'),
+      ...(this.state.isCopied && { visible: true })
+    };
+
+    return (
+      <Tooltip {...tooltipProps}>
+        <Button
+          className={classNames('no-select', this.props.className)}
```
gregaubert: Not related to your sprint, but I just noticed that this `.no-select` class is only used in those clipboard buttons but is declared in misc in sonar-web, could we move it to `buttons.css` in sonar-ui-common ?

components/controls/ClipboardIconButton.tsx
```diff
@@ -0,0 +1,64 @@
...
+          data-clipboard-text={this.props.copyValue}
+          innerRef={node => (this.copyButton = node)}>
+          {this.props.children || translate('copy')}
```
gregaubert: This one should directly render the copy icon, and not take children in parameter, neither try to render the `translate('copy')`

components/controls/ClipboardIconButton.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  render() {
+    const tooltipProps = {
+      overlay: translate(this.state.isCopied ? 'copied_action' : 'copy_to_clipboard'),
+      ...(this.state.isCopied && { visible: true })
+    };
+
+    return (
+      <Tooltip {...tooltipProps}>
+        <Button
```
gregaubert: Instead of using tooltip and button and introducing a new kind/style of icon buttons, I think we should use the `ButtonIcon`. It has a tooltip prop too.

components/controls/ClipboardIconButton.tsx
```diff
@@ -0,0 +1,64 @@
...
+interface State {
+  isCopied: boolean;
+}
+
+export default class ClipboardIconButton extends ClipboardBase {
+  copyButton?: HTMLElement | null;
+  mounted = false;
+  state: State = { isCopied: false };
+
+  showTooltip = () => {
```
gregaubert: And if you use the `ButtonIcon` this definitely won't handle showing the tooltip or not. So let's make it more generic and rename it here and in the base to something like `handleSuccessCopy`.
In this component all we will do is change the text and revert it 1sec later, and leave all the display aspect of the tooltip to `ButtonIcon`.

components/controls/ClipboardIconButton.tsx
gregaubert: Ok but currently the `ButtonIcon` tooltip is broken because of the change with the `ThemeContext`, I'll push a small fix to it.

components/controls/ClipboardIconButton.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { Button } from './buttons';
+import ClipboardBase from './ClipboardBase';
+import { translate } from '../../helpers/l10n';
+
+interface State {
+  isCopied: boolean;
+}
+
+export default class ClipboardIconButton extends ClipboardBase {
+  copyButton?: HTMLElement | null;
```
gregaubert: This should not be declared here, it's already done in the `ClipboardBase` component.

components/controls/ClipboardIconButton.tsx
```diff
@@ -0,0 +1,64 @@
...
+  mounted = false;
```
gregaubert: This should not be declared here, it's already done in the `ClipboardBase` component.

components/controls/Tooltip.tsx
```diff
@@ -105,7 +105,7 @@ export class TooltipInner extends React.Component<Props, State> {
       this.addEventListeners();
     } else if (
       // closes
-      (this.props.visible === false && prevProps.visible === true) ||
+      (!this.props.visible && prevProps.visible === true) ||
```
gregaubert, line 108: I think we don't need those changes anymore if we use `ButtonIcon`

components/controls/ClipboardBase.tsx
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  componentWillUnmount() {
+    this.mounted = false;
+    if (this.clipboard) {
+      this.clipboard.destroy();
+    }
+  }
+
+  showTooltip: ShowTooltipFunction = () => {
```
gregaubert: Like I said on another comment, lets rename this

components/controls/ClipboardBase.tsx
```diff
@@ -0,0 +1,67 @@
...
+import * as Clipboard from 'clipboard';
+
+interface Props {
+  className?: string;
+  copyValue: string;
+}
+
+type ShowTooltipFunction = (() => never) | VoidFunction;
+
+export default class ClipboardBase extends React.PureComponent<Props> {
```
gregaubert: I guess it's ok for now but I'm not a huge fan of having this component exported out of sonar-ui-common, since it's not usable on it's own...
Maybe a solution to this could be to do one `clipboard.tsx` file that would contain this base component (not exported) and the two other components that depend on it and would be exported.

gregaubert (changes requested):
Not ready for me yet. Mainly I think we should use `ButtonIcon`, see my comments.

components/controls/ClipboardBase.tsx
wouter-admiraal-sonarsource: I agree with Greg. The approach we have is using [HOCs](https://reactjs.org/docs/higher-order-components.html). We don't have any HOC yet in sonar-ui-common, but this could be the first one.

Check out `withScrollTo()` in `sonar-web`. I think it's the closest to what we try to achieve here (compared to the others, which are mostly about application state).

components/controls/ClipboardButton.tsx
```diff
@@ -19,52 +19,20 @@
...
-    }
-  }
-
-  componentWillUnmount() {
-    this.mounted = false;
-    if (this.clipboard) {
-      this.clipboard.destroy();
-    }
-  }
-
```
wouter-admiraal-sonarsource, line 67: So if we go the HOC approach, all this will be moved to the HOC.

wouter-admiraal-sonarsource (changes requested):
Sorry, I approved without noticing the implementation completely changed. My bad. I'm not a big fan of this new approach. See my comments below.

siegfried-ehret-sonarsource:
@wouter-admiraal-sonarsource @gregaubert can you review that again ?

It is using render props and not an HOC, TypeScript was mean with me (and Jay, who tried to save me).

components/controls/Tooltip.tsx
```diff
@@ -96,7 +96,7 @@ export class TooltipInner extends React.Component<Props, State> {
   componentDidUpdate(prevProps: Props, prevState: State) {
     if (
       // opens
-      (this.props.visible === true && !prevProps.visible) ||
+      (this.props.visible === true && prevProps.visible === false) ||
```
siegfried-ehret-sonarsource: @gregaubert no, we need this.

gregaubert (commented):
I just finished my review, figured it would be faster to just apply my comments than write them. So here are two commits, I let you have a look and ping me to discuss what needs to be discussed :)

I also pushed a matching commit in your branch on sonar-enterprise.

gregaubert (approved):

siegfried-ehret-sonarsource:
It's green but I want to show where we are now to Matthieu before merging.
Also, I can't build sonar-entreprise from https://github.com/SonarSource/sonar-enterprise/pull/1932 because of some 💀 💥 :rage4: webpack bundle size error.

gregaubert:
> Also, I can't build sonar-entreprise from [SonarSource/sonar-enterprise#1932](https://github.com/SonarSource/sonar-enterprise/pull/1932) because of some   :rage4: webpack bundle size error.

You have this issue when trying to build locally ? It might be because you try to build while linking sonar-ui-common...

siegfried-ehret-sonarsource:
Matthieu validated the thing ! 🎉

PR #13, opened 2019-07-18
SONAR-11714 Remove .badge from Select list CSS

wouter-admiraal-sonarsource:
Ahem, sorry... Should have linked locally... The new design doesn't work well with these styles.

Relates to https://github.com/SonarSource/sonar-enterprise/pull/1930

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,10 @@ language: node_js
 sudo: false
 install: true
 jdk: oraclejdk11
-node_js: "10.16.0"
+node_js: "10.16"
 dist: trusty
 
-before_install: if [[ `yarn -v` != 1.17* ]]; then curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.17.3 && export PATH="$HOME/.yarn/bin:$PATH"; fi
+before_install: if [[ `yarn -v` != 1.15* ]]; then curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.15.2 && export PATH="$HOME/.yarn/bin:$PATH"; fi
 
 env:
   - TARGET=BUILD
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -2,11 +2,9 @@ language: node_js
 sudo: false
 install: true
 jdk: oraclejdk11
-node_js: "10.16.0"
+node_js: "8"
```
gregaubert: To align with `sonar-enterprise` it should be `10.15.3`, not `8`

.travis.yml
```diff
@@ -2,11 +2,9 @@ language: node_js
...
 dist: trusty
 
-before_install: if [[ `yarn -v` != 1.17* ]]; then curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.17.3 && export PATH="$HOME/.yarn/bin:$PATH"; fi
```
gregaubert, line 8: I don't know what the default version is on Travis, let's use `1.15.2` to align with `sonar-enterprise`.
Or update `build.gradle` from `sonar-enterprise` to use a more recent one.

.travis.yml
```diff
@@ -2,10 +2,10 @@ language: node_js
 sudo: false
 install: true
 jdk: oraclejdk11
-node_js: "10.16.0"
+node_js: "10.15.3"
```
siegfried-ehret-sonarsource: @wouter-admiraal-sonarsource 10.16.0 (current LTS) is not available on Travis ?

wouter-admiraal-sonarsource: We align it on sonar-enterprise, which uses 10.15.3.

siegfried-ehret-sonarsource: Shouldn't we align the other way and have the latest LTS ?

wouter-admiraal-sonarsource: We could, although we don't care that much (IMHO). It's only for building, not running any application itself. And I realize we're speaking of the `.travis.yml` file, which yes, we can def. set to the LTS 😅 . But we can keep 10.15.3 in `package.json`.

gregaubert (approved):

PR #16, opened 2019-07-30
Sort typescript import

philippe-perrin-sonarsource:


package.json
```diff
@@ -143,4 +143,4 @@
     "printWidth": 100,
     "singleQuote": true
   }
-}
+}
```
gregaubert: We are just missing the EOF empty line

gregaubert (approved):
Looks good! :) Thanks!

package.json
philippe-perrin-sonarsource: Damn it

PR #20, opened 2019-08-06
Hardening

philippe-perrin-sonarsource:
See:

https://jira.sonarsource.com/browse/SONAR-12322 (Original [PR](https://github.com/SonarSource/sonar-enterprise/pull/1644) / [JIRA](https://jira.sonarsource.com/browse/SC-669))
**https://jira.sonarsource.com/browse/SONAR-11742**

See also:

Sonar-Enterprise PR : https://github.com/SonarSource/sonar-enterprise/pull/1989

**Checklist**

* [x] Write/update ITs
* [x] Functional validation
* [ ] ~Documentation update~
* [x] Update changelog


components/controls/FavoriteBase.tsx
```diff
@@ -48,6 +48,12 @@ export default class FavoriteBase extends React.PureComponent<Props, State> {
     this.mounted = true;
   }
 
+  componentDidUpdate(prevProps: Props) {
+    if (prevProps.favorite !== this.props.favorite) {
+      this.setState({ favorite: this.props.favorite });
```
gregaubert: Good catch, @siegfried-ehret-sonarsource do you remember why this component has a state ?
To me it seems like this component should not have a state and only be controlled by it's parent...

philippe-perrin-sonarsource: I agree. It might looks like this. It solves my problem and does not break the initial fix (stars disappearing when scrolling). The state could be hold by Favorite component which is the only one who calls FavoriteBase and which is called by all the component that use the favorite function (if I'm not mistaking).

```typescript
export default class FavoriteBase extends React.PureComponent<Props> {
  toggleFavorite = () =>
    this.props.favorite ? this.props.removeFavorite() : this.props.addFavorite();

  render() {
    const { favorite } = this.props;
    const tooltip = favorite
      ? translate('favorite.current', this.props.qualifier)
      : translate('favorite.check', this.props.qualifier);
    const ariaLabel = translate('favorite.action', favorite ? 'remove' : 'add');
    return (
      <Tooltip overlay={tooltip}>
        <ButtonLink
          aria-label={ariaLabel}
          className={classNames('favorite-link', 'link-no-underline', this.props.className)}
          onClick={this.toggleFavorite}>
          <FavoriteIcon favorite={favorite} />
        </ButtonLink>
      </Tooltip>
    );
  }
}

```

gregaubert: Yeah it would simplify things so much... I think the initial fix went crazy, unless I don't remember something about this state ?
Let's wait for Siegfried input.

siegfried-ehret-sonarsource: I don't remember :(
If the star doesn't disappear on scroll I am happy with everything.

siegfried-ehret-sonarsource: BTW the Favorite component should be in sonar-ui-common no ?

philippe-perrin-sonarsource: > BTW the Favorite component should be in sonar-ui-common no ?

Hmm, I thought so at first, but in fact it is the one responsible for the api call (add/remove). So now I do think it should exists on the sonar-web side. In fact, I think that it is one of the 2 reasons to exists : wrap those calls & and hold state.

gregaubert: Indeed it seems to only exist to wrap those calls, and we don't have access to api methods in sonar-ui-common. So either we drop it or keep it in sonar-web.

wouter-admiraal-sonarsource: If I recall correctly, the star state disappeared when scrolling the Projects page. It probably had something to do with the virtualized list we use there. Rings a bell @siegfried-ehret-sonarsource ?

gregaubert (approved):
Looking good! Simplification FTW!

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-
+SONAR-12322 Project Mark/Unmark as favorite button is not refreshing properly when switching of project
```
wouter-admiraal-sonarsource: Don't forget the bullet point 🙂 .

components/controls/FavoriteBase.tsx
```diff
@@ -25,75 +25,26 @@ import { ButtonLink } from './buttons';
...
-      () => {
-        if (this.mounted) {
-          this.setState({ favorite: false });
-        }
-      },
-      () => {}
-    );
-  }
-
+export default class FavoriteBase extends React.PureComponent<Props> {
```
wouter-admiraal-sonarsource: Crazy thought: should we rename this to `FavoriteButton` or something? I find `FavoriteBase` makes it look like an abstract class.

components/controls/FavoriteBase.tsx
```diff
@@ -25,75 +25,26 @@ import { ButtonLink } from './buttons';
...
   render() {
-    const { favorite } = this.state;
+    const { favorite } = this.props;
     const tooltip = favorite
       ? translate('favorite.current', this.props.qualifier)
       : translate('favorite.check', this.props.qualifier);
```
wouter-admiraal-sonarsource: Really nitpicking here, but as you're deconstructing `props` anyway, might was well add `qualifier` to the list 😉 .

wouter-admiraal-sonarsource (commented):
Looks good! I'll validate functionally a bit later, stay tuned.

wouter-admiraal-sonarsource (approved):
Works! (combined with your PR on sonar-enterprise)

philippe-perrin-sonarsource:
Sorry for the noise guys. But I would like to avoid packaging too much version of the lib. So I am grouping my changes. Next time I'll wait for a more stable version before asking for review.

philippe-perrin-sonarsource:
Just need an extra validation on [that](https://jira.sonarsource.com/browse/SONAR-11742) and I'll be ready. BTW, smallest commit eva !

wouter-admiraal-sonarsource (approved):
LGTM

gregaubert (approved):
Looks good!

PR #22, opened 2019-08-13
Update icon for ClipboardIconButton

siegfried-ehret-sonarsource:
Needed for MMF-1781, as a starting point to revamp the copy icon for SonarCloud. See icons at the end of points 1 and 2 in this mockup screen:

FirstAnalysisTutorial_3 2-BitBucket--create-pipeline-step2

See: 

* https://jira.sonarsource.com/browse/MMF-1781

**Checklist**

* [x] Write/update ITs
* [ ] Functional validation
* [ ] Documentation update
* [x] Update changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Export new request methods: get, getText, parseText
+- Update the icon for ClipboardIconButton
 
 ## 0.0.20
 
--- a/components/icons/CopyIcon.tsx
+++ b/components/icons/CopyIcon.tsx
@@ -23,11 +23,11 @@ import Icon, { IconProps } from './Icon';
 export default function CopyIcon({ className, fill = 'currentColor', size }: IconProps) {
   return (
     <Icon className={className} size={size}>
-      <path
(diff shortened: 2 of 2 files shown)


gregaubert (commented):
Looks good, just one question though: should we really allow to customize ? or just change the icon altogether? If we allow to customize, it means we can end up with lot of different visual for the copy icon button, is it really something we want to allow ?

siegfried-ehret-sonarsource:
WDYT @wouter-admiraal-sonarsource @jeremy-davis-sonarsource @philippe-perrin-sonarsource ?
The existing icon was given by @mathieu-cutivel-sonarsource.

gregaubert:
@laurawacrenier @mathieu-cutivel-sonarsource is it really important to have a different copy icon for SQ and SC ? Or can you guys agree on one of them that we use for both products ?

laurawacrenier:
> @laurawacrenier @mathieu-cutivel-sonarsource is it really important to have a different copy icon for SQ and SC ? Or can you guys agree on one of them that we use for both products ?

No, we should use the same. I decided to change it because the first one was not following our current icon style for action icons (should be solid, not outlined). 

@mathieu-cutivel-sonarsource it looks like there is a flaw in our design review process because I never saw this icon before and was very surprised to find something like that already existed. Let's talk about it when you're back.

siegfried-ehret-sonarsource:
Thanks @laurawacrenier, I'll update it.

gregaubert:
Thanks for you input @laurawacrenier ! So let's change the icon altogether and not make the button customizable! :)

gregaubert (approved):
:+1:

PR #23, opened 2019-08-16
Styled-component

philippe-perrin-sonarsource:
**Checklist**

* [x] Write/update ITs
* [x] Functional validation
* [ ] ~Public documentation update~
* [x] Private documentation update
* [x] Ping a member of the other team (SQ/SC)

This is the associated SQ [PR](https://github.com/SonarSource/sonar-enterprise/pull/2007).


philippe-perrin-sonarsource:
@wouter-admiraal-sonarsource & @siegfried-ehret-sonarsource : You might want to have a look at it.

components/ui/Alert.tsx
```diff
@@ -19,47 +19,143 @@
...
+export interface Props {
   display?: AlertDisplay;
   variant: AlertVariant;
 }
 
-export function Alert(props: AlertProps & React.HTMLAttributes<HTMLDivElement>) {
+export function Alert(props: Props & React.HTMLAttributes<HTMLDivElement>) {
+  const isInline = () => props.display === 'inline';
+  const isBanner = () => props.display === 'banner';
+  const getAlertVariantInfo = (theme: SonarUiCommonTheme) => {
```
siegfried-ehret-sonarsource: This could be extracted from the `function Alert` component, adding the variant as a parameter.

components/ui/Alert.tsx
```diff
@@ -19,47 +19,143 @@
...
+          borderColor: theme.colors.alertBorderInfo,
+          backGroundColor: theme.colors.alertBackgroundInfo
+        };
+      }
+      default:
+        throw new Error(`Unsupported alert variant : ${props.variant}`);
+    }
+  };
+
+  const AlertIcon = styled.div.attrs(p => ({ variantInfo: getAlertVariantInfo(p.theme) }))`
```
siegfried-ehret-sonarsource: What happens if you make a typo in the property name ? Do you have a warning or something ?

components/ui/Alert.tsx
```diff
@@ -19,47 +19,143 @@
...
+    justify-content: center;
+    align-items: center;
+    width: calc(${isBanner() ? 2 : 4} * ${p.theme.sizes.gridSize});
+    ${!isBanner() ? 'border-right: 1px solid;' : ''};
+    border-color: ${p.variantInfo.borderColor};
+    `}
+  `;
+
+  const AlertContent = styled.div`
+    ${(p: any) => `
```
siegfried-ehret-sonarsource: I find this hard to understand :sob:. What is the meaning of `p` ?

components/test-utils.tsx
```diff
@@ -0,0 +1,31 @@
...
+import * as React from 'react';
+import { SonarUiCommonThemeProvider, Theme } from './ThemeContext';
+import { MockedTheme } from './__mocks__/theme';
+
+export function shallowWithTheme(children: JSX.Element, theme: Theme = MockedTheme) {
+  return shallow(<SonarUiCommonThemeProvider theme={theme}>{children}</SonarUiCommonThemeProvider>)
+    .find('ContextConsumer')
+    .dive()
+    .find('ContextProvider');
+}
```
wouter-admiraal-sonarsource: Shouldn't this be put in `testUtils.ts` instead?

components/ui/Alert.tsx
```diff
@@ -19,47 +19,143 @@
...
-import ThemeContext from '../ThemeContext';
-import './Alert.css';
+import { SonarUiCommonThemeConsumer, Theme } from '../ThemeContext';
 
-type AlertDisplay = 'banner' | 'block' | 'inline';
+type AlertDisplay = 'banner' | 'inline' | 'block';
 type AlertVariant = 'error' | 'warning' | 'success' | 'info';
 
-export interface AlertProps {
+export interface Props {
```
wouter-admiraal-sonarsource: I believe we now prefix all exported `Props` interfaces with the component name, to make it easier to find them. You should keep it as `AlertProps`.

components/ui/Alert.tsx
```diff
@@ -19,47 +19,143 @@
...
+  };
+
+  const AlertIcon = styled.div.attrs(p => ({ variantInfo: getAlertVariantInfo(p.theme) }))`
+    ${(p: any) => `
+    flex: 0 0 auto;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    width: calc(${isBanner() ? 2 : 4} * ${p.theme.sizes.gridSize});
+    ${!isBanner() ? 'border-right: 1px solid;' : ''};
```
wouter-admiraal-sonarsource: Is there a way to pass this `isBanner()` return value as a prop/attr to the styled component? If so, we could move the declaration of these "internal" components (`AlertIcon`, `AlertContent` and `AlertInner`) outside the render logic of `Alert` itself.

components/ui/Alert.tsx
philippe-perrin-sonarsource: I was not aware of such things. But that makes sense.

components/ui/Alert.tsx
philippe-perrin-sonarsource: p is the props of the styled-components, which contains the global styled-component theme. See 2 lines below.

philippe-perrin-sonarsource:
This typescript [plugin](https://github.com/Microsoft/typescript-styled-plugin#configuration) combined with the extension for [ VS Code](https://github.com/styled-components/vscode-styled-components) or [IntelliJ](https://github.com/styled-components/webstorm-styled-components) brings us intellisense and linting support.

components/ui/Alert.tsx
philippe-perrin-sonarsource: Not really no. The goal is to extend the component props. But props remains of type `ThemedStyledProps<>`.

See
https://www.styled-components.com/docs/basics#attaching-additional-props
https://www.styled-components.com/docs/api#attrs

philippe-perrin-sonarsource: Actually there is a way to type the additional props and then to have typo feedback. See fixup commit.

components/ui/Alert.tsx
philippe-perrin-sonarsource: See fixup commit

components/ui/Alert.tsx
philippe-perrin-sonarsource: See fixup commit

components/ui/Alert.tsx
```diff
@@ -25,41 +25,162 @@ import AlertErrorIcon from '../icons/AlertErrorIcon';
...
+            className={classNames('alert', className)}
+            isInline={isInline}
+            role="alert"
+            variantInfo={variantInfo}
+            {...domProps}>
+            <StyledAlertInner isBanner={isBanner}>
+              <Tooltip overlay={translate('alert.tooltip', variant)}>
+                <StyledAlertIcon isBanner={isBanner} variantInfo={variantInfo}>
+                  <SonarUiCommonThemeConsumer>
+                    {theme => getAlertVariantInfo(theme, variant).icon}
```
jeremy-davis-sonarsource: Why not use `variantInfo`? You don't need to repeat the `ThemeConsumer` either.

components/ui/Alert.tsx
```diff
@@ -25,41 +25,162 @@ import AlertErrorIcon from '../icons/AlertErrorIcon';
...
+
+    a,
+    .button-link {
+      border-color: ${props.theme.colors.darkBlue};
+    }
+  `}
+`;
+
+function getAlertVariantInfo(theme: SonarUiCommonTheme, variant: string) {
+  switch (variant) {
```
jeremy-davis-sonarsource: How about a hashmap instead of the switch, it's easier to read IMO, and no default required if you type `variant` as `AlertVariant`

components/ui/Alert.tsx
```diff
@@ -25,41 +25,162 @@ import AlertErrorIcon from '../icons/AlertErrorIcon';
...
 
+interface VariantInformation {
+  icon: JSX.Element;
+  color: string;
+  borderColor: string;
+  backGroundColor: string;
+}
+
+const StyledAlertIcon = styled.div<{ isBanner: boolean; variantInfo: VariantInformation }>`
+  ${props => css`
```
jeremy-davis-sonarsource: you don't need the `css` function

components/ui/Alert.tsx
```diff
@@ -25,41 +25,162 @@ import AlertErrorIcon from '../icons/AlertErrorIcon';
...
+    flex: 0 0 auto;
+    display: flex;
+    justify-content: center;
+    align-items: center;
```
jeremy-davis-sonarsource: All these can be put outside of the prop-dependent body, and they will be part of a "base" class. The variable css will then be in other additional classes

components/ui/Alert.tsx
```diff
@@ -25,41 +25,162 @@ import AlertErrorIcon from '../icons/AlertErrorIcon';
...
+      ? css`
+          border-right: 1px solid;
+        `
+      : ''};
+    border-color: ${props.variantInfo.borderColor};
+  `}
+`;
+
+const StyledAlertContent = styled.div`
+  ${props => css`
```
jeremy-davis-sonarsource: No need for `css`

components/ui/Alert.tsx
```diff
@@ -25,41 +25,162 @@ import AlertErrorIcon from '../icons/AlertErrorIcon';
...
+    flex: 1 1 auto;
+    padding: ${props.theme.sizes.gridSize} calc(2 * ${props.theme.sizes.gridSize});
+    overflow: auto;
+    text-align: left;
+  `}
+`;
+
+const StyledAlertInner = styled.div<{ isBanner: boolean }>`
+  ${props => css`
```
jeremy-davis-sonarsource: No need for `css`

components/ui/Alert.tsx
```diff
@@ -25,41 +25,162 @@ import AlertErrorIcon from '../icons/AlertErrorIcon';
...
+    display: flex;
+    align-items: stretch;
+
+    ${props.isBanner
+      ? css`
```
jeremy-davis-sonarsource: No need for `css`

components/ui/Alert.tsx
```diff
@@ -25,41 +25,162 @@ import AlertErrorIcon from '../icons/AlertErrorIcon';
...
+
+const StyledAlertIcon = styled.div<{ isBanner: boolean; variantInfo: VariantInformation }>`
+  ${props => css`
+    flex: 0 0 auto;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    width: calc(${props.isBanner ? 2 : 4} * ${props.theme.sizes.gridSize});
+    ${!props.isBanner
+      ? css`
```
jeremy-davis-sonarsource: No need for `css`

components/ui/Alert.tsx
```diff
@@ -25,41 +25,162 @@ import AlertErrorIcon from '../icons/AlertErrorIcon';
...
+          padding-left: ${props.theme.sizes.pagePadding};
+          padding-right: ${props.theme.sizes.pagePadding};
+          box-sizing: border-box;
+        `
+      : ''}
+  `}
+`;
+
+const StyledAlert = styled.div<{ isInline: boolean; variantInfo: VariantInformation }>`
+  ${props => css`
```
jeremy-davis-sonarsource: no need for `css`

components/ui/Alert.tsx
```diff
@@ -25,41 +25,162 @@ import AlertErrorIcon from '../icons/AlertErrorIcon';
...
+    margin-bottom: ${props.theme.sizes.gridSize};
+    border: 1px solid;
+    border-radius: 2px;
+    border-color: ${props.variantInfo.borderColor};
+    background-color: ${props.variantInfo.backGroundColor};
+    color: ${props.variantInfo.color};
+
+    ${props.isInline
+      ? css`
```
jeremy-davis-sonarsource: no need for `css`

jeremy-davis-sonarsource (commented):
Looks pretty good! 
You don't need the `css` function anywhere, it's just a helper to interpolate prop functions. within the scope of `styled`, you're covered.

philippe-perrin-sonarsource:
> Looks pretty good!
> You don't need the `css` function anywhere, it's just a helper to interpolate prop functions. within the scope of `styled`, you're covered.

I know, but it helps triggering the highlighting of the css. I think I should find another way.

package.json
```diff
@@ -48,14 +49,16 @@
...
     "@types/react-virtualized": "9.21.0",
+    "@types/styled-components": "4.1.8",
     "@typescript-eslint/parser": "1.11.0",
     "cpy-cli": "2.0.0",
     "date-fns": "1.30.1",
     "diff": "^4.0.1",
     "enzyme": "3.10.0",
     "enzyme-adapter-react-16": "1.14.0",
-    "enzyme-to-json": "3.3.5",
+    "enzyme-to-json": "^3.4.0",
```
gregaubert: Fixed versions should be used

components/theme.ts
```diff
@@ -0,0 +1,65 @@
...
+(ThemeContext as any)['_currentValue2'] = defautTheme;
+const typedStyled = styled as ThemedBaseStyledInterface<SonarUiCommonTheme>;
+const typedCss = css as BaseThemedCssFunction<SonarUiCommonTheme>;
+const typedContext = ThemeContext as React.Context<SonarUiCommonTheme>;
+const typedProvider = ThemeProvider as ThemeProviderComponent<SonarUiCommonTheme>;
+const typedConsumer = ThemeConsumer as React.Consumer<SonarUiCommonTheme>;
+
+export {
+  typedProvider as SonarUiCommonThemeProvider,
+  typedConsumer as SonarUiCommonThemeConsumer,
```
gregaubert: Could we just export as `ThemeProvider` and `ThemeConsumer` to have shorter names ?  We already import them from `sonar-ui-common/components/theme` so we know it comes from sonar-ui-common

gregaubert (commented):
Regarding the build issue, like you suggested I think we should move `lodash` as a peer dependency of `sonar-ui-common`, because I saw that in fact it's also imported in another chunk from the node modules of `sonar-ui-common`.
But this change alone won't allow you to build `sonar-web` under the chunk size limit (see my comment on the other PR)

components/theme.ts
```diff
@@ -0,0 +1,65 @@
...
+export interface Theme {
+  colors: T.Dict<string>;
+  sizes: T.Dict<string>;
+  rawSizes: T.Dict<number>;
+  fonts: T.Dict<string>;
+  zIndexes: T.Dict<string>;
+  others: T.Dict<string>;
+}
+
+const defautTheme: Theme = {
```
wouter-admiraal-sonarsource: NP: typo in const name.

```suggestion
const defaultTheme: Theme = {
```

components/ui/Alert.tsx
```diff
@@ -25,41 +25,155 @@ import AlertErrorIcon from '../icons/AlertErrorIcon';
...
+      : ''}
+
+    a,
+    .button-link {
+      border-color: ${props.theme.colors.darkBlue};
+    }
+  `}
+`;
+
+function getAlertVariantInfo(theme: Theme, variant: string): AlertVariantInformation {
```
wouter-admiraal-sonarsource: NP: `variant` can be typed as `AlertVariant`.

wouter-admiraal-sonarsource: NP bis: you could deconstruct `theme` to `{ colors }` for improved readability.

wouter-admiraal-sonarsource (approved):
Great job! Really nice implem 👍 !

jeremy-davis-sonarsource (changes requested):
As discussed, to get rid of the extra `css` calls and have better class generation, we'll follow the suggested way of handling variable styles and inline all `${props => props.etc...}`

philippe-perrin-sonarsource:
@wouter-admiraal-sonarsource You might to take a look at the last fixup commit. According to @jeremy-davis-sonarsource comments, I have rewrite some style to avoid having to many `css` helper.

jeremy-davis-sonarsource (approved):
LGTM :+1: Nice work!

PR #25, opened 2019-08-21
SONAR-12380 Tooltips should not be displayed outside the viewport

wouter-admiraal-sonarsource:
See:

* https://jira.sonarsource.com/browse/SONAR-12380

**Checklist**

* [ ] ~Write/update ITs~
* [x] Functional validation
* [ ] ~Documentation update~ 
* [x] Update changelog
* [ ] ~Fix `DocTooltip` (if possible)~
* [ ] ~Styled Components? Depends on https://github.com/SonarSource/sonar-ui-common/pull/23~

### DIFF
--- a/components/controls/Tooltip.tsx
+++ b/components/controls/Tooltip.tsx
@@ -20,7 +20,8 @@
 import { throttle } from 'lodash';
 import * as React from 'react';
 import { createPortal, findDOMNode } from 'react-dom';
-import ScreePositionFixer from './ScreenPositionFixer';
+import ThemeContext from '../ThemeContext';
+import ScreenPositionFixer from './ScreenPositionFixer';
 import './Tooltip.css';
 
 export type Placement = 'bottom' | 'right' | 'left' | 'top';
@@ -45,10 +46,19 @@ interface Measurements {
 }
 
 interface OwnState {
+  flipped: boolean;
+  placement?: Placement;
   visible: boolean;
 }
 
 type State = OwnState & Partial<Measurements>;
+
+const FLIP_MAP: { [key in Placement]: Placement } = {
+  left: 'right',
+  right: 'left',
+  top: 'bottom',
+  bottom: 'top'
+};
 
 function isMeasured(state: State): state is OwnState & Measurements {
   return state.height !== undefined;
@@ -77,9 +87,13 @@ export class TooltipInner extends React.Component<TooltipProps, State> {
     mouseEnterDelay: 0.1
   };
 
+  static contextType = ThemeContext;
+
   constructor(props: TooltipProps) {
     super(props);
     this.state = {
+      flipped: false,
+      placement: props.placement,
       visible: props.visible !== undefined ? props.visible : false
     };
     this.throttledPositionTooltip = throttle(this.positionTooltip, 10);
@@ -94,6 +108,14 @@ export class TooltipInner extends React.Component<TooltipProps, State> {
   }
 
   componentDidUpdate(prevProps: TooltipProps, prevState: State) {
+    if (this.props.placement !== prevProps.placement) {
+      this.setState({ placement: this.props.placement });
+      // Break. This will trigger a new componentDidUpdate() call, so the below
+      // positionTooltip() call will be correct. Otherwise, it might not use
+      // the new state.placement value.
+      return;
+    }
+
     if (
       // opens
       (this.props.visible === true && !prevProps.visible) ||
@@ -141,11 +163,24 @@ export class TooltipInner extends React.Component<TooltipProps, State> {
   };
 
   getPlacement = (): Placement => {
-    return this.props.placement || 'bottom';
+    return this.state.placement || 'bottom';
   };
 
   tooltipNodeRef = (node: HTMLElement | null) => {
     this.tooltipNode = node;
+  };
+
+  adjustArrowPosition = (
+    placement: Placement,
+    { leftFix, topFix }: { leftFix: number; topFix: number }
+  ) => {
+    switch (placement) {
+      case 'left':
+      case 'right':
+        return { marginTop: -topFix };
+      default:
+        return { marginLeft: -leftFix };
+    }
   };
 
   positionTooltip = () => {
@@ -186,22 +221,23 @@ export class TooltipInner extends React.Component<TooltipProps, State> {
 
       // save width and height (and later set in `render`) to avoid resizing the tooltip element,
       // when it's placed close to the window edge
-      const measurements: Measurements = {
+      this.setState({
         left: window.pageXOffset + left,
         top: window.pageYOffset + top,
         width,
         height
-      };
-      this.setState(measurements);
+      });
     }
   };
 
   clearPosition = () => {
     this.setState({
+      flipped: false,
       left: undefined,
       top: undefined,
       width: undefined,
-      height: undefined
+      height: undefined,
+      placement: this.props.placement
     });
   };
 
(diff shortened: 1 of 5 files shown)


components/controls/Tooltip.tsx
```diff
@@ -186,12 +185,56 @@ export class TooltipInner extends React.Component<TooltipProps, State> {
...
+      return { placement: 'left', style: { ...style, left } };
+    } else if (placement === 'bottom' && topFix && topFix < -threshold) {
+      const { top = style.top } = this.getPosition('top') || {};
+      return { placement: 'top', style: { ...style, top } };
+    } else if (placement === 'top' && topFix && topFix > threshold) {
+      const { top = style.top } = this.getPosition('bottom') || {};
+      return { placement: 'bottom', style: { ...style, top } };
+    } else {
+      return { placement, style };
+    }
```
philippe-perrin-sonarsource: As discussed, in order to reduce the pain of reading this block, you might want to use an "opposite" lookup table and test threshold against absolute value. Just ideas.

components/controls/Tooltip.tsx
```diff
@@ -262,35 +305,47 @@ export class TooltipInner extends React.Component<TooltipProps, State> {
...
+                  : undefined;
+
+                if (style && !this.props.forcePlacement) {
+                  const adjustment = this.adjustTooltipPosition(placement, style, {
+                    leftFix,
+                    topFix
+                  });
+                  style = adjustment.style;
+                  placement = adjustment.placement;
+                }
```
philippe-perrin-sonarsource: Also, as discussed, extracting the logic from the render function might be great for readability.

philippe-perrin-sonarsource (commented):
Some discussion feedback for now. Also, isn't it the good time to isolate css from that component ? Styled-components or whatever (depending on what we want to use).

components/controls/Tooltip.tsx
```diff
@@ -157,6 +155,7 @@ export class TooltipInner extends React.Component<TooltipProps, State> {
     // eslint-disable-next-line react/no-find-dom-node
     const toggleNode = findDOMNode(this);
 
+    let measurements;
```
gregaubert: Can we type as `Measurement | undefined` ? So that the return type of the method is correct too ?

components/controls/Tooltip.tsx
gregaubert: Pretty hard to digest indeed

gregaubert (approved):
It's looking good, I tested a few tooltips but I guess the best way to be sure everything is fine is to merge and dogfood the change.

components/controls/Tooltip.tsx
gregaubert: This block is pretty hard to digest indeed, it's nice if you have a way to make it more readable!

wouter-admiraal-sonarsource:
Completely changed the implementation. We don't re-compute in `render()`, but instead update the state, which triggers a re-render. It's slightly less "OK", as we update the state during the rendering (\*slap hand\*), _but_ it makes it much easier to understand, and we keep the rendering (mostly) logic-free.

@philippe-perrin-sonarsource About Styled Components, as long as your PR isn't merged, and we're not 100% sure it's the road we want to take, I prefer not updating it here.

components/controls/Tooltip.tsx
```diff
@@ -252,8 +286,85 @@ export class TooltipInner extends React.Component<TooltipProps, State> {
...
+            // re-compute our positioning.
+            height: undefined,
+            placement: FLIP_MAP[placement]
+          }),
+          () => {
+            if (this.state.visible) {
+              // Force a re-positioning, as "only" updating the state doesn't
+              // recompute the position, only re-renders with the previous
+              // position (which is no longer correct).
+              this.positionTooltip();
```
gregaubert, line 330: This will also call a `setState`, so the there will be two renders triggered from this ? for the first `setState` that updates the `flipped` and then the one for the `positionTooltip`, is it the two renders you are talking about in `needsFlipping` ?

gregaubert: React can't batch those `setState` because you use the callback to trigger `positionTooltip` (since it requires the placement to be up to date in the state.
But could pass the new placement directly to `positionTooltip` to avoid using the callback and maybe that would allow react to batch those `setState` and only render once ?

wouter-admiraal-sonarsource: Good catch! No, this wouldn't trigger 2 re-renders, but (potentially) **4** (because of `ScreenPositionFixer`)!

I will call `setState()` only once 👍 .

gregaubert: No need to call it only once, it might be difficult to put in place ? Just allowing to pass the new placement to positionTooltip so it doesn't have to rely on the state could be enough, then react would be able to batch those state change and trigger only one render.

components/controls/Tooltip.tsx
```diff
@@ -318,3 +403,5 @@ class TooltipPortal extends React.Component {
     return createPortal(this.props.children, this.el);
   }
 }
+
+TooltipInner.contextType = ThemeContext;
```
philippe-perrin-sonarsource: Private thoughts, but i really dislike this kind of things. Visually, it is outside of the context of the class and is a weird things to read. I think we should use the other proposed style for that. 

```ts
class MyClass extends React.Component {
  static contextType = MyContext;
  render() {
    let value = this.context;
    /* render something based on the value */
  }
}
```

What do you think ?

philippe-perrin-sonarsource: https://reactjs.org/docs/context.html#classcontexttype

components/controls/Tooltip.tsx
```diff
@@ -318,3 +403,5 @@ class TooltipPortal extends React.Component {
     return createPortal(this.props.children, this.el);
   }
```
philippe-perrin-sonarsource: Could we move it to its own file ? Would it be wise ? Readability | SRP ...

components/controls/Tooltip.tsx
wouter-admiraal-sonarsource: Hmm, I realize it won't work. We need to explicitly set `height` to undefined, which triggers a re-render of `ScreenPositionFixer`, which is then ready to re-compute the positioning once more.

For a non-flipping Tooltip, the flow looks like this (not counting the initial rendering of the toggle node itself):

Screenshot 2019-08-23 at 14 03 30

For a flipping Tooltip, it looks like this:

Screenshot 2019-08-23 at 14 04 06

So, by default, **without** my fix, a Tooltip renders twice when shown, and twice when hidden (total of 4 times).
**With** my fix, a Tooltip renders four times when shown, and twice when hidden (a total of 6 times).

I think I might be able to cut this to three times, at a slight cost of readability. But is the ROI high enough, considering:

1. This will happen pretty rarely
2. We're "only" re-rendering a Tooltip, not a whole section of the application. Is winning a single render call, in rare situations, really worth that extra effort and "cost" in maintainability?

components/controls/Tooltip.tsx
```diff
@@ -20,7 +20,8 @@
 import { throttle } from 'lodash';
 import * as React from 'react';
 import { createPortal, findDOMNode } from 'react-dom';
-import ScreePositionFixer from './ScreenPositionFixer';
+import ThemeContext from '../ThemeContext';
+import ScreenPositionFixer from './ScreenPositionFixer';
 import './Tooltip.css';
 
 export type Placement = 'bottom' | 'right' | 'left' | 'top';
```
philippe-perrin-sonarsource, line 27: I know this is not part of the rework done here, but I'd rather see an enum here, so we avoid spreading string literals around the code. It is of course subject to discussion.

components/controls/Tooltip.tsx
philippe-perrin-sonarsource: Also this syntaxt is already used in the class.

```ts
  static defaultProps = {
    mouseEnterDelay: 0.1
  };

```

components/controls/Tooltip.tsx
```diff
@@ -252,8 +286,85 @@ export class TooltipInner extends React.Component<TooltipProps, State> {
     this.handleMouseLeave();
   };
 
-  render() {
+  needsFlipping = (leftFix: number, topFix: number) => {
+    // We can live with a tooltip that's slightly positioned over the toggle
+    // node. Only trigger if it really starts overlapping, as the re-positioning
+    // is quite expensive, needing 2 re-renders.
+    const threshold = this.context.theme.rawSizes.grid;
```
philippe-perrin-sonarsource, line 295: I think this threshold might be a bit too big.

tooltip_threshol

philippe-perrin-sonarsource (commented):
Great job reworking this complex component. What about the tooltip that will need to be move both on x & y axis ?

tooltips_bottom_left

components/controls/Tooltip.tsx
gregaubert: Ok thanks for the details, indeed I don't think it's worth investing more into this, let's keep your current implementation! :+1:

components/controls/Tooltip.tsx
wouter-admiraal-sonarsource: What would we move to its own file?

philippe-perrin-sonarsource: The `TooltipPortal` class

wouter-admiraal-sonarsource: I prefer keeping it here. It should _never_ be used on its own. Keeping it part of this file at least leaves it scoped, and prevents ambiguity.

components/controls/Tooltip.tsx
wouter-admiraal-sonarsource: Hmm, on second thought, although I'm in favor of this in principle, this component is used in so many places, we would need to refactor many, many calls, for very little gain. I'd leave it like this for now.

components/controls/Tooltip.tsx
wouter-admiraal-sonarsource: Ha! That's a `HelpToolTip`, they lazy load their content 😑 . So the height is not correctly computed, and the positioning is off.

philippe-perrin-sonarsource (approved):
As discussed, it would be great if lazy-loaded content could be display gracefully, Still, great works !

PR #26, opened 2019-08-26
SONAR-12414 Add new ClickEventBoundary to prevent click event bubbling

wouter-admiraal-sonarsource:
See:

* https://jira.sonarsource.com/browse/SONAR-12414
  In issues, when we interact with text inside a dropdown (like the changelog, or the comments), selecting text, and releasing the mouse button, will open the issue. This new component serves as a "boundary" after which a click event cannot bubble further.
  See https://github.com/SonarSource/sonar-enterprise/pull/2025

**Checklist**

* [ ] Write/update ITs
* [x] Functional validation
* [ ] ~Documentation update~
* [x] Update changelog

### DIFF
--- a/components/controls/ClickEventBoundary.tsx
+++ b/components/controls/ClickEventBoundary.tsx
@@ -1,0 +1,35 @@
+/*
+ * Sonar UI Common
+ * Copyright (C) 2019-2019 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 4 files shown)


siegfried-ehret-sonarsource (approved):

components/controls/ClickEventBoundary.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export interface ClickEventBoundaryProps {
+  children: React.ReactElement;
+}
+
+export default function ClickEventBoundary({ children }: ClickEventBoundaryProps) {
+  return React.cloneElement(children, {
+    onClick: (e: React.SyntheticEvent<MouseEvent>) => {
+      e.stopPropagation();
+      if (children.props.onClick !== undefined) {
```
philippe-perrin-sonarsource: You might want to do that to be sure not to call a onClick prop that is not defined (null / undefined / not a function).

```suggestion
      if (typeof children.props.onClick === 'function') {
```

philippe-perrin-sonarsource (approved):
LGTM

components/controls/ClickEventBoundary.tsx
wouter-admiraal-sonarsource: Good catch 👍

PR #27, opened 2019-08-26
SONAR-12372 Fix Radio preventing a modal from closing on hitting Esc

wouter-admiraal-sonarsource:
See:

* https://jira.sonarsource.com/browse/SONAR-12372
  If a Modal contains Radio buttons (see "Add Condition" modal under Quality Gates), you can no longer close it using the Esc key (confirmed in FF and Chrome, on macOS). For some reason, the way we `blur()` prevents the keydown event from firing. We don't really need it here, so I just removed the whole event handling altogether.
  See https://github.com/SonarSource/sonar-enterprise/pull/2025

**Checklist**

* [ ] ~Write/update ITs~
* [x] Functional validation
* [ ] ~Documentation update~
* [x] Update changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- SONAR-12372 Fix Radio preventing a Modal from closing using the Esc key
+
 ## 0.0.22
 
 - Fix positioning logic in ScreenPositionFixer
--- a/components/controls/Radio.tsx
+++ b/components/controls/Radio.tsx
@@ -29,9 +29,7 @@ interface Props {
 }
 
 export default class Radio extends React.PureComponent<Props> {
-  handleClick = (event: React.MouseEvent<HTMLAnchorElement>) => {
-    event.preventDefault();
-    event.currentTarget.blur();
+  handleClick = () => {
     this.props.onCheck(this.props.value);
   };
 


siegfried-ehret-sonarsource (commented):
Is there a way to unit test that change ?

wouter-admiraal-sonarsource:
> Is there a way to unit test that change ?

Good question. Not sure... It didn't affect Safari, for instance, so it's not something that might be reproducible in any Node environment. Furthermore, DOM event bubbling is hard to test in UTs (because it's completely dependent on the "fake" DOM, so could be inaccurate). I don't know if it's worth the effort?

philippe-perrin-sonarsource:
From what I understood, after the blur is triggered, the body becomes the focused element. So the esc key will be trigger on it, where there is no esc key event handler (I suppose it is located on the modal itself). Focusing on an input makes it the source of the next esc key event, which bubbles up to the modal esc key event handler.

This behavior does not surprise me, but do you know why we trigger the blur event in the first place ?

radio

wouter-admiraal-sonarsource:
> From what I understood, after the blur is triggered, the body becomes the focused element. So the esc key will be trigger on it, where there is no esc key event handler (I suppose it is located on the modal itself). Focusing on an input makes it the source of the next esc key event, which bubbles up to the modal esc key event handler.

I hadn't even realised this! I thought the keyup event handler was on the document, but you're right; it's probably on the modal itself.

> but do you know why we trigger the blur event in the first place ?

I imagine it was originally because of the `:focus` styles. Clicking on a radio would keep it focused, showing a blue border around it in certain browsers. But this style is now explicitly disabled for most form elements, so it no longer applies.

philippe-perrin-sonarsource (approved):
As discussed, LGTM.

gregaubert:
@wouter-admiraal-sonarsource Can we re-discuss this fix ? Since we dropped the `event.preventDefault()` now the url is always polluted with `#` when clicking a radio button (I noticed this because it breaks the downgrade feedback page from SC). Can we only drop the `event.currentTarget.blur();` instead ?

wouter-admiraal-sonarsource:
@gregaubert Whoopsy daisy 😅 . Sure thing! The culprit was really the `blur()` call. I didn't imagine the `preventDefault()` added any value (not the case in most UIs I tested), so I removed it. But we can def. add it again. You want to? Shall I?

gregaubert:
Ok thanks for the clarification! :+1: 
I'll do it, I just wanted to be sure the culprit was only `blur()`

PR #34, opened 2019-09-04
Move billing extension into the core

gregaubert:
I'm currently moving our billing extension directly into the sonarcloud core, and this default function is needed, but I clearly don't want to also depend on react-select inside sonar-web so I'm exposing it here. I know it's not ideal since it's kind of only for sonarcloud and it's billing part, but I still think it's better than also depending on react-select in the core just for that.

**Checklist**

* [ ] ~Write/update ITs~
* [x] Functional validation
* [ ] ~Documentation update~
* [x] Update changelog

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "sonar-ui-common",
-  "version": "0.0.25",
+  "version": "0.0.26",
   "description": "Common UI lib for SonarQube and SonarCloud",
   "repository": "SonarSource/sonar-ui-common",
   "license": "LGPL-3.0",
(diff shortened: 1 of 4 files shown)


jeremy-davis-sonarsource (approved):
LGTM

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "sonar-ui-common",
-  "version": "0.0.25",
+  "version": "0.0.26",
```
jeremy-davis-sonarsource, line 3: Looking forward to v1!

PR #36, opened 2019-09-04
New LazyLoading function using React.Lazy, Suspense and ErrorBoundaries

gregaubert:
See:

New lazy loading function using more recent stuff from react. I implemented this mainly because it correctly handles the typings unlike our old lazyloading function that doesn't. And I needed a good handling of typing for my move of billing into sonar-web.
So then we can slowly replace old lazyloading use with this new one, I'm not sure we need to really replace all of them at once, since it's working well currently.

**Checklist**

* [x] Write/update ITs
* [x] Functional validation
* [ ] Documentation update
* [x] Update changelog

### DIFF
--- a/components/lazyLoadComponent.tsx
+++ b/components/lazyLoadComponent.tsx
@@ -1,0 +1,73 @@
+/*
+ * Sonar UI Common
+ * Copyright (C) 2019-2019 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import * as React from 'react';
+import { translate } from '../helpers/l10n';
+import { requestTryAndRepeatUntil } from '../helpers/request';
+import { Alert } from './ui/Alert';
+
+interface ImportError {
+  request?: string;
+}
+
+export function lazyLoadComponent<T extends React.ComponentType<any>>(
+  factory: () => Promise<{ default: T }>,
+  displayName?: string
+) {
+  const LazyComponent = React.lazy(() =>
+    requestTryAndRepeatUntil(factory, { max: 1, slowThreshold: 2 }, () => true)
+  );
+
+  function LazyComponentWrapper(props: React.ComponentProps<T>) {
+    return (
+      <LazyErrorBoundary>
+        <React.Suspense fallback={null}>
+          <LazyComponent {...props} />
+        </React.Suspense>
+      </LazyErrorBoundary>
+    );
+  }
+
+  LazyComponentWrapper.displayName = displayName;
+  return LazyComponentWrapper;
+}
+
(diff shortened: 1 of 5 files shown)


components/__tests__/reactLazyLoad-test.tsx
```diff
@@ -0,0 +1,55 @@
...
+  expect(wrapper.render()).toMatchSnapshot();
+  expect(factory).toBeCalledTimes(1);
+});
+
+it('should correctly handle import errors', () => {
+  const factory = jest.fn().mockImplementation(() => import('../controls/Checkbox'));
+  const LazyComponent = reactLazyLoad(factory);
+  const wrapper = mount(<LazyComponent />);
+  wrapper.find('Suspense').simulateError({ request: 'test' });
+  expect(wrapper.find('Alert').exists()).toBe(true);
```
jeremy-davis-sonarsource: :fire:
Do I get a duck point? :grin:

components/reactLazyLoad.tsx
```diff
@@ -0,0 +1,73 @@
...
+import * as React from 'react';
+import { translate } from '../helpers/l10n';
+import { requestTryAndRepeatUntil } from '../helpers/request';
+import { Alert } from './ui/Alert';
+
+interface ImportError {
+  request?: string;
+}
+
+export function reactLazyLoad<T extends React.ComponentType<any>>(
```
jeremy-davis-sonarsource: How about `lazyLoadComponent`, I'm weirdly bothered by `reactLazyLoad`

jeremy-davis-sonarsource (approved):
I'm bothered by the name, but it looks good otherwise!

components/__tests__/reactLazyLoad-test.tsx
gregaubert: You sure do!

components/reactLazyLoad.tsx
gregaubert: Yeah I didn't put too much thought into the name, started with `newLazyLoad` xD
I'm fine with `lazyLoadComponent` :+1:

components/lazyLoadComponent.tsx
```diff
@@ -0,0 +1,73 @@
...
+  displayName?: string
+) {
+  const LazyComponent = React.lazy(() =>
+    requestTryAndRepeatUntil(factory, { max: 1, slowThreshold: 2 }, () => true)
+  );
+
+  function LazyComponentWrapper(props: React.ComponentProps<T>) {
+    return (
+      <LazyErrorBoundary>
+        <React.Suspense fallback={null}>
```
siegfried-ehret-sonarsource, line 40: Should we make the `fallback` configurable for customization depending on the usage ?

components/lazyLoadComponent.tsx
```diff
@@ -0,0 +1,73 @@
...
+          <LazyComponent {...props} />
+        </React.Suspense>
+      </LazyErrorBoundary>
+    );
+  }
+
+  LazyComponentWrapper.displayName = displayName;
```
siegfried-ehret-sonarsource, line 47: It would be nice to have a default display name maybe ?

siegfried-ehret-sonarsource (approved):
Very nice ! 2 mini things, let me know what you think.

components/lazyLoadComponent.tsx
gregaubert: I was wondering about that too... It could be a good idea and at the same time I don't know if we want to end up with lot of different behavior with each dev providing their own fallback...

I'm thinking that maybe when we will want to have a real fallback displaying something while the component is loading we will update this component and every lazy loaded component will have the same ?

components/lazyLoadComponent.tsx
gregaubert: The default display name will be `LazyComponentWrapper` if the `displayName` is not passed as a second argument.

components/lazyLoadComponent.tsx
siegfried-ehret-sonarsource: Depending on the type of component we may not have the same spinner for example ? I don't know... We can do that when we need it.

gregaubert: Yeah maybe, I say let's do that when there is a need.

PR #37, opened 2019-09-10
Fix import

siegfried-ehret-sonarsource:


gregaubert (approved):

gregaubert:
I approved but I think we don't need to put this in the changelog file

PR #38, opened 2019-09-11
New loading alert type

gregaubert:
**Checklist**

* [ ] ~Write/update ITs~
* [x] Functional validation
* [ ] ~Documentation update~
* [x] Update changelog

### DIFF
--- a/components/ui/__tests__/__snapshots__/Alert-test.tsx.snap
+++ b/components/ui/__tests__/__snapshots__/Alert-test.tsx.snap
@@ -429,6 +429,77 @@ exports[`should render basic alert 4`] = `
 </div>
 `;
 
+exports[`should render basic alert 5`] = `
+.c2 {
+  -webkit-flex: 0 0 auto;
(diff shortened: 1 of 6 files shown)


components/ui/__tests__/__snapshots__/Alert-test.tsx.snap
```diff
@@ -429,6 +429,77 @@ exports[`should render basic alert 4`] = `
...
+<div
+  class="alert alert-test c0"
+  id="error-message"
+  role="alert"
+>
+  <div
+    class="c1"
+  >
+    <div
+      class="c2"
```
gregaubert, line 492: This is a bit shitty, we don't see the spinner in the snapshot... It's because I use a DeferredSpinner... But I don't plan to fix this snapshot here, I started another PR to have a real Spinner styled component and the snapshot will look better.

siegfried-ehret-sonarsource (approved):

PR #39, opened 2019-09-17
Add KeyCodes enum

wouter-admiraal-sonarsource:
See:

* [This comment from Philippe](https://github.com/SonarSource/sonar-enterprise/pull/2041#discussion_r324669188)

**Checklist**

* [ ] ~Write/update ITs~
* [ ] ~Functional validation~
* [ ] ~Documentation update~
* [x] Update changelog

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,7 +103,8 @@
   "jest": {
     "coverageDirectory": "<rootDir>/build/coverage",
     "collectCoverageFrom": [
-      "{components,helpers}/**/*.{ts,tsx,js}"
+      "{components,helpers}/**/*.{ts,tsx,js}",
+      "!helpers/{keycodes,testUtils}.{ts,tsx}"
     ],
     "coverageReporters": [
       "lcovonly",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Provide an enum with keycodes, to simplify handling keyboard events.
+
 ## 0.0.28
 
 - SC-937 Add new loading variant for Alert messages
--- a/helpers/keycodes.ts
+++ b/helpers/keycodes.ts
@@ -1,0 +1,68 @@
+/*
+ * Sonar UI Common
+ * Copyright (C) 2019-2019 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+export enum KeyCodes {
+  LeftArrow = 37,
+  UpArrow = 38,
+  RightArrow = 39,
+  DownArrow = 40,
+
+  Alt = 18,
+  Backspace = 8,
+  CapsLock = 20,
+  Command = 93,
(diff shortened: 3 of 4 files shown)


philippe-perrin-sonarsource (approved):
LGTM, although I would rename the file with a more obvious name (keycodes.ts ?), WDYT ?. Thanks for doing that. I didn't take the time to do it before my holidays ... :-)

wouter-admiraal-sonarsource:
@philippe-perrin-sonarsource Hmm, I just did some double checking, and not all codes are correct. I guess it can depend on your keyboard layout and OS. I will remove all the ones that could potentially be incorrect, and only keep those of which we can be certain.

philippe-perrin-sonarsource:
Damn, yeah I didn't check them. I took the first example I found on github. At least, those that we need are correct. Maybe we should start with them only : enter, arrowup, arrowdown.

wouter-admiraal-sonarsource:
Done. Much shorter, but we probably don't need most of the original list. I think this should suffice.

wouter-admiraal-sonarsource:
@philippe-perrin-sonarsource @gregaubert As you're both on another OS, could you check if those codes are indeed correct?

Just typing this in your browser's console should do the trick 🙂 :
```
document.addEventListener('keyup', (e) => {
  const codes = {
  37: 'LeftArrow',
  38: 'UpArrow',
  39: 'RightArrow',
  40: 'DownArrow',

  18: 'Alt',
  8: 'Backspace',
  20: 'CapsLock',
  93:'Command',
  17: 'Ctrl',
  46: 'Delete',
  35: 'End',
  13: 'Enter',
  27: 'Escape',
  36: 'Home',
  34: 'PageDown',
  33: 'PageUp',
  16: 'Shift',
  32: 'Space',
  9: 'Tab',

  48: 'Zero',
  49: 'One',
  50: 'Two',
  51: 'Three',
  52: 'Four',
  53: 'Five',
  54: 'Six',
  55: 'Seven',
  56: 'Eight',
  57: 'Nine',

  65: 'A',
  66: 'B',
  67: 'C',
  68: 'D',
  69: 'E',
  70: 'F',
  71: 'G',
  72: 'H',
  73: 'I',
  74: 'J',
  75: 'K',
  76: 'L',
  77: 'M',
  78: 'N',
  79: 'O',
  80: 'P',
  81: 'Q',
  82: 'R',
  83: 'S',
  84: 'T',
  85: 'U',
  86: 'V',
  87: 'W',
  88: 'X',
  89: 'Y',
  90: 'Z',

  112: 'F1',
  113: 'F2',
  114: 'F3',
  115: 'F4',
  116: 'F5',
  117: 'F6',
  118: 'F7',
  119: 'F8',
  120: 'F9',
  121: 'F10',
  122: 'F11',
  123: 'F12'
}
  console.log(e.which, 'You just typed ', codes[e.which], ', correct?');
});
```

philippe-perrin-sonarsource (approved):
LG on windows.

wouter-admiraal-sonarsource:
I've added a new commit, which will ignore coverage of `helpers/testUtils.tsx` and `helpers/keycodes.ts`. They're not supposed to be tested, so we can also ignore their coverage. WDYT?

package.json
```diff
@@ -103,7 +103,8 @@
   "jest": {
     "coverageDirectory": "<rootDir>/build/coverage",
     "collectCoverageFrom": [
-      "{components,helpers}/**/*.{ts,tsx,js}"
+      "{components,helpers}/**/*.{ts,tsx,js}",
+      "!helpers/{keycodes,testUtils}.{ts,tsx}"
```
gregaubert, line 107: You might want to add a `sonar.coverage.exclusions` propriety to the `sonar-project.properties` files too. With those files referenced.

gregaubert (commented):
Looks good. Just a few points I noticed on Linux:
* Numbers from the numpad are not the same as those at the top of the letters
* Maybe we should not reference the F keys ? We currently don't use them in the app and most of theme have effects on the browser directly so they can't be used in the app.

wouter-admiraal-sonarsource:
> Looks good. Just a few points I noticed on Linux:
> 
> * Numbers from the numpad are not the same as those at the top of the letters
> * Maybe we should not reference the F keys ? We currently don't use them in the app and most of theme have effects on the browser directly so they can't be used in the app.

* Aha! Didn't check those. I'll remove them. We don't really need them for now.
* Indeed. I'll remove them as well.

gregaubert (approved):

PR #40, opened 2019-09-17
Fix button-link transitions

jeremy-davis-sonarsource:
A general fix for a problem that was spotted while validating [PR 2071](https://github.com/SonarSource/sonar-enterprise/pull/2071)


gregaubert (approved):
Thanks!

PR #41, opened 2019-09-19
Switch to @emotion 10 instead of styled-components

gregaubert:
Two main reasons for that switch:

* smaller library
* avoid conflicts with atlaskit in sonar-bitbucketcloud

Also the `vscode-styled-component` plugin works fine with emotion.

**Checklist**

* [x] Write/update ITs
* [x] Functional validation
* [ ] ~Documentation update~
* [x] Update changelog


CHANGELOG.md
```diff
@@ -2,7 +2,10 @@
 
 ## Unreleased
 
+## 0.0.29
+
 - Provide an enum with keycodes, to simplify handling keyboard events.
+- SC-1098 Switch to @emotion instead of styled-components
```
jeremy-davis-sonarsource, line 9: Can you please add `Fix button-link transitions` ? It's from my commit and I forgot to add it to the Changelog!

package.json
```diff
@@ -66,27 +70,24 @@
...
     "typescript": "3.5.2",
-    "typescript-styled-plugin": "0.14.0",
     "whatwg-fetch": "3.0.0"
   },
   "scripts": {
     "clean": "rm -rf build/dist",
     "build": "scripts/build.sh",
     "package": "yarn build && cd build/dist && yarn pack",
     "test": "jest",
-    "test-coverage": "jest --coverage",
```
philippe-perrin-sonarsource, line 89: Why removing this line ?

helpers/testUtils.tsx
```diff
@@ -163,12 +160,3 @@ export async function waitAndUpdate(wrapper: ShallowWrapper<any, any> | ReactWra
...
 }
-
-export function renderWithTheme(toRender: JSX.Element, theme?: Theme) {
-  return shallow(<ThemeProvider theme={{ ...mockedTheme, ...theme }}>{toRender}</ThemeProvider>)
-    .find('ContextConsumer')
-    .dive()
-    .find('ContextProvider')
-    .dive()
-    .render();
-}
```
philippe-perrin-sonarsource, line 174: If I am not mistaking, this file originally had an `.ts` extension which I changed to `.tsx` in order to support the `renderWithTheme` method. Maybe you should set it back to `.ts`.

package.json
```diff
@@ -27,20 +27,24 @@
...
   },
   "peerDependencies": {
+    "@emotion/core": "10.0.17",
+    "@emotion/styled": "10.0.17",
+    "emotion-theming": "10.0.19",
     "lodash": "4.17.15",
     "react": "16.8.6",
-    "react-dom": "16.8.6",
-    "styled-components": "4.3.2"
+    "react-dom": "16.8.6"
```
philippe-perrin-sonarsource: Not related to that PR, but as these are `peerDependencies`, I think we should be less restrictive in the allowed version.

https://docs.npmjs.com/about-semantic-versioning#using-semantic-versioning-to-specify-update-types-your-package-can-accept

I would say we should at least allows patch releases to be used.

components/ui/Alert.tsx
```diff
@@ -60,13 +60,13 @@ const StyledAlertContent = styled.div`
   padding: ${props => props.theme.sizes.gridSize} calc(2 * ${props => props.theme.sizes.gridSize});
 `;
 
-const alertInnerIsBannerMixin = css`
-  min-width: ${props => props.theme.sizes.minPageWidth};
-  max-width: ${props => props.theme.sizes.maxPageWidth};
+const alertInnerIsBannerMixin = (props: ThemeProps) => css`
```
jeremy-davis-sonarsource: Deconstruct `props`' `theme`?

components/theme.ts
```diff
@@ -17,15 +17,17 @@
...
-  css,
-  ThemeConsumer,
-  ThemeContext,
-  ThemedBaseStyledInterface,
-  ThemeProvider,
-  ThemeProviderComponent
-} from 'styled-components';
+import { css, ThemeContext as EmotionThemeContext } from '@emotion/core';
+// eslint-disable-next-line import/no-named-default
+import { CreateStyled, default as emotionStyled } from '@emotion/styled';
```
philippe-perrin-sonarsource: What about ?

```suggestion
// Remove the eslint-disable comments
import emotionStyled, { CreateStyled } from '@emotion/styled';
```

components/theme.ts
```diff
@@ -36,30 +38,26 @@ export interface Theme {
...
-  zIndexes: {},
-  others: {}
-};
+export interface ThemeProps {
+  theme: Theme;
+}
+
+const ThemeContext = EmotionThemeContext as React.Context<Theme>;
 
 // Hack : override the default value of the context used for theme by styled-component
```
philippe-perrin-sonarsource:
```suggestion
// Hack : override the default value of the context used for theme by emotion
```

jeremy-davis-sonarsource (approved):
LGTM! Emotion ftw!

components/theme.ts
```diff
@@ -36,30 +38,26 @@ export interface Theme {
...
 
-const defaultTheme: Theme = {
-  colors: {},
-  sizes: {},
-  rawSizes: {},
-  fonts: {},
-  zIndexes: {},
-  others: {}
-};
+export interface ThemeProps {
```
philippe-perrin-sonarsource: I'm not sure to see why you'll need that interface. I would drop it and rewrite the Alert like it : 

```ts
const alertInnerIsBannerMixin = (theme: Theme) => css`
  min-width: ${theme.sizes.minPageWidth};
  max-width: ${theme.sizes.maxPageWidth};
  margin-left: auto;
  margin-right: auto;
  padding-left: ${theme.sizes.pagePadding};
  padding-right: ${theme.sizes.pagePadding};
  box-sizing: border-box;
`;

const StyledAlertInner = styled.div<{ isBanner: boolean }>`
  display: flex;
  align-items: stretch;
  ${props => (props.isBanner ? alertInnerIsBannerMixin(props.theme) : null)}
`;
```

WDYT ?

components/ui/Alert.tsx
philippe-perrin-sonarsource: See my comment on `Theme.ts`

components/ui/__tests__/Alert-test.tsx
```diff
@@ -17,29 +17,36 @@
...
+    expect(wrapper.prop('variantInfo')).toMatchSnapshot();
   });
 });
 
 it('should render inline alert', () => {
-  expect(render({ display: 'inline' })).toMatchSnapshot();
+  expect(shallowRender({ display: 'inline' }).render()).toMatchSnapshot();
 });
 
 it('should render banner alert', () => {
```
philippe-perrin-sonarsource, line 40: Although I understand why this test is interesting, I think it is redundant with the snapshot one. As there is a css class linked to this banner mode, the snapshot already does tests it. Also, you don't test the inline mode, which is basically the same thing. The only purpose of those 2 flags is to adjust css which appears in the snapshot. Therefore they are tested IMO. WDYT ?

philippe-perrin-sonarsource (commented):
Minor comments here and there. Great job reworking the whole theme thing in the context of tests. And thank you for taking time to migrate to @emotion.

package.json
gregaubert: Because of all our projects it's the only one that has a `test-coverage` script, and it's not used.

helpers/testUtils.tsx
gregaubert: Good catch!

package.json
gregaubert: Agree!

components/theme.ts
gregaubert: I guess in this case indeed it doesn't make much sens to have this type. This is because we use the mixin conditionally. But for another component where you would use a mixin without condition juste like this: `${myMixin}` it's easier to have the type available, because I don't want to have to write `${ props => myMixin(props.theme)}` in this case.

So I think we should keep the theme and try keep a consistent way of writing mixins.

philippe-perrin-sonarsource: Fair point. Then I think `ThemeProps` is misleading. It is not the props of theme, it a props with a theme. What about `StyledProps` or `ThemedProps` ?

philippe-perrin-sonarsource (approved):

components/theme.ts
gregaubert: Indeed, I'll rename `ThemedProps`

gregaubert:
@philippe-perrin-sonarsource Can you have a look at my last commit?

philippe-perrin-sonarsource (approved):
LGTM

PR #42, opened 2019-09-20
Update RadioToggle CSS

siegfried-ehret-sonarsource:
Use a pointer instead of the usual arrow to show that it is clickable.

### DIFF
--- a/components/controls/RadioToggle.css
+++ b/components/controls/RadioToggle.css
@@ -52,6 +52,7 @@
   color: var(--darkBlue);
   height: calc(var(--controlHeight) - 2px);
   line-height: calc(var(--controlHeight) - 2px);
+  cursor: pointer;
 }
 
 .radio-toggle input[type='radio'] {


siegfried-ehret-sonarsource:
@gregaubert not sure if this is worth mentioning in the changelog...

philippe-perrin-sonarsource (approved):
I would add, does it even deserves to be a PR ? 😆

components/controls/RadioToggle.css
```diff
@@ -52,6 +52,7 @@
   color: var(--darkBlue);
   height: calc(var(--controlHeight) - 2px);
   line-height: calc(var(--controlHeight) - 2px);
+  cursor: pointer;
```
gregaubert, line 55: Are you really sure it's not `pointer: cursor;` ? :trollface:

gregaubert (approved):
It looks ok, but given the huge number of changes and the complexity of the review we can never be 100% sure, so I'll just trust you on this one!

PR #52, opened 2019-10-16
Update README.md

philippe-perrin-sonarsource:


README.md
```diff
@@ -31,11 +31,11 @@ It will provide syntax highlighting, quick-fix and intellisense.
 
 ## What to improve
 
-- [ ] CSS of component based on theme (we currently rely on the fact that all usage of sonar-ui-common also compile CSS of components with the correct CSS custom properties setup)
+- [ ] Add a global index file
+- [ ] Expose sonar-ui-common from sonar-web to reduce bundle size from extensions, currently it's duplicated in every extensions. Also it should help with theming and allows us to remove most of the call to ThemeProvider.
```
gregaubert: Can we also add something like: "Our current way of using `ThemeProvider` everywhere is mostly due to the fact that every extension is loading a different `ThemeProvider` component (it means that this `ThemeProvider` is duplicated for each extension)."
So that we remember our conclusions about this subject.

gregaubert (approved):

PR #54, opened 2019-10-18
SC-1242 Invert code smell icon

gregaubert:
See:

* One more small adjustment to SC-1242, so that the code smell icon looks more like a hazard logo : 

And less like a cinema magnetic tape:

**Checklist**

* [x] Update changelog


wouter-admiraal-sonarsource (approved):
LGTM!

Although I much prefer running into a cinema reel than an area marked with a nuclear hazard symbol 🤔 ...

PR #58, opened 2019-10-22
Add automatic release script

gregaubert:
A little script to ease the release process.
@wouter-admiraal-sonarsource Can you check if it works on MacOs ? And if not suggest update ?
@philippe-perrin-sonarsource Like we discussed already, you don't necessarily need to check now if you don't want to. Just don't expect it to work on your first try when you will need it and you will have to update it at that time if necessary.

:warning:  **WARNING**: If you want to try it, first comment the publish and push commands

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -6,15 +6,24 @@ if [ $# -eq 0 ]
     exit 1
 fi
 
+if [ ! $(which npmrc) ]
+  then
+    echo "You need to install npmrc and configure it correctly, check the readme. Call:"
+    echo "  npm install -g npmrc"
+    exit 1
+fi
+
 git stash
-sed -i "s/## Unreleased/## Unreleased\n\r## $1/g" CHANGELOG.md
-sed -i "s/\"version\": \".*\",/\"version\": \"$1\",/g" package.json
+sed "s/## Unreleased/## Unreleased__NEW_LINE__## $1/g" CHANGELOG.md | awk '{ sub(/__NEW_LINE__/,"\n\n"); print }' > CHANGELOG.md.back
+mv CHANGELOG.md.back CHANGELOG.md
+sed "s/\"version\": \".*\",/\"version\": \"$1\",/g" package.json > package.json.back
+mv package.json.back package.json
 yarn
 yarn package
+npmrc npm
 git add CHANGELOG.md package.json
 git commit -m "Prepare version $1"
 git tag $1
-npmrc npm
 yarn publish ./build/dist/sonar-ui-common-v$1.tgz
 npmrc default
 git push
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+
+if [ $# -eq 0 ]
+  then
+    echo "Missing argument for the new version to release"
+    exit 1
+fi
+
+git stash
+sed -i "s/## Unreleased/## Unreleased\n\r## $1/g" CHANGELOG.md
+sed -i "s/\"version\": \".*\",/\"version\": \"$1\",/g" package.json
+yarn
+yarn package
+git add CHANGELOG.md package.json
+git commit -m "Prepare version $1"
+git tag $1
+npmrc npm
+yarn publish ./build/dist/sonar-ui-common-v$1.tgz
+npmrc default
+git push
+git push origin $1
+git stash pop
--- a/README.md
+++ b/README.md
@@ -71,6 +71,24 @@ A good practice before publishing the package is to make sure everything work we
 
 ## Publish new version
 
+### Automated way
+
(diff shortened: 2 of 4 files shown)


scripts/release.sh
```diff
@@ -0,0 +1,22 @@
...
+git stash
+sed -i "s/## Unreleased/## Unreleased\n\r## $1/g" CHANGELOG.md
+sed -i "s/\"version\": \".*\",/\"version\": \"$1\",/g" package.json
+yarn
+yarn package
+git add CHANGELOG.md package.json
+git commit -m "Prepare version $1"
+git tag $1
+npmrc npm
+yarn publish ./build/dist/sonar-ui-common-v$1.tgz
```
gregaubert, line 27: If you want to try locally without really doing a release, comment this line

scripts/release.sh
```diff
@@ -0,0 +1,22 @@
...
+npmrc default
+git push
```
gregaubert: If you want to try locally without really doing a release, comment this

scripts/release.sh
```diff
@@ -0,0 +1,22 @@
...
+git push origin $1
```
gregaubert: If you want to try locally without really doing a release, comment this

gregaubert, line 30: If you want to try locally without really doing a release, comment those two lines

scripts/release.sh
wouter-admiraal-sonarsource: **Careful!!** Also comment the `git stash` one 😅 !!

wouter-admiraal-sonarsource: Or even better, create a local commit with these lines commented out 😉 .

scripts/release.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+if [ $# -eq 0 ]
+  then
+    echo "Missing argument for the new version to release"
+    exit 1
+fi
+
```
wouter-admiraal-sonarsource, line 8: I would add the following to make sure you don't run the script without `npmrc` installed (because the whole script will run regardless, and this might have unexpected consequences):

```bash
if [ ! $(which npmrc) ]
  then
    echo "You need to install npmrc. Call:"
    echo "  npm install -g npmrc"
    exit 1
fi
```

scripts/release.sh
gregaubert: git stash is ok no ? it will unstash at the end...

scripts/release.sh
```diff
@@ -0,0 +1,22 @@
...
+git stash
+sed -i "s/## Unreleased/## Unreleased\n\r## $1/g" CHANGELOG.md
+sed -i "s/\"version\": \".*\",/\"version\": \"$1\",/g" package.json
```
wouter-admiraal-sonarsource: A few things:

* On macOS, you need to provide an argument to `-i`, which can be an empty string (`-i ""`) (BSD style).
* I believe we only use Linux line endings (`\n`) in our `README.md` and `CHANGELOG.md` files (I see a `^M` in there BTW; we should be careful with that). I would use `\n\n` instead of `\r\n` (notice I suggest **two** empty lines instead of just one, because that's the pattern we used for all releases in the `CHANGELOG.md` file).
* Finally, the newlines don't work with `sed` in BSD 😑 (hence, not on macOS either). Perhaps we can use something like this:
  ```bash
  sed "s/## Unreleased/## Unreleased__NEW_LINE__## $1/g" CHANGELOG.md | awk '{ sub(/__NEW_LINE__/,"\n\n"); print }' > CHANGELOG.md.back
  mv CHANGELOG.md.back CHANGELOG.md
   ```
  (redirect to `CHANGELOG.md.back`, as you're normally not allowed to update a file you're reading from in one go).

Putting all this together:

```bash
sed "s/## Unreleased/## Unreleased__NEW_LINE__## $1/g" CHANGELOG.md | awk '{ sub(/__NEW_LINE__/,"\n\n"); print }' > CHANGELOG.md.back
mv CHANGELOG.md.back CHANGELOG.md
sed -i "" "s/\"version\": \".*\",/\"version\": \"$1\",/g" package.json
```

wouter-admiraal-sonarsource (approved):

scripts/release.sh
wouter-admiraal-sonarsource: `git stash` will stash the lines you commented out. So it will perform the tasks you commented.

gregaubert: ohhh good catch! ^^ it was ok on my machine because didn't have the file in git yet! That was close one!

wouter-admiraal-sonarsource: Unfortunately I didn't catch it before running it. But when I saw the output of `yarn publish ...` I frantically `^C`'d 😄. It's OK, it didn't publish anything.

scripts/release.sh
gregaubert: Thanks for the suggestions! :)

gregaubert: I couldn't make it work on linux with the (`-i ""`) so I did it with a back file like in the first sed.

PR #63, opened 2019-11-14
SONAR-12636 Add a disabled state to the toggle component

philippe-perrin-sonarsource:
Add a disabled state to the toggle component, in order to avoid multiple click when calling an api for example.

Sonar-enterprise associated [PR](https://github.com/SonarSource/sonar-enterprise/pull/2243).

### DIFF
--- a/components/controls/Toggle.tsx
+++ b/components/controls/Toggle.tsx
@@ -24,6 +24,7 @@ import { Button } from './buttons';
 import './Toggle.css';
 
 interface Props {
+  disabled?: boolean;
   name?: string;
   onChange?: (value: boolean) => void;
   value: boolean | string;
@@ -43,11 +44,12 @@ export default class Toggle extends React.PureComponent<Props> {
   };
 
   render() {
+    const { disabled, name } = this.props;
     const value = this.getValue();
     const className = classNames('boolean-toggle', { 'boolean-toggle-on': value });
 
     return (
-      <Button className={className} name={this.props.name} onClick={this.handleClick}>
+      <Button className={className} disabled={disabled} name={name} onClick={this.handleClick}>
         <div className="boolean-toggle-handle">
           <CheckIcon size={12} />
         </div>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,13 @@
 
 ## Unreleased
 
+## 0.0.37
+
 - Bump to typescript@3.6.4
 - Fix BoxedTabs alignment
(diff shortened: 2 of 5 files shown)


jeremy-davis-sonarsource (commented):
Looks all good, but I question the solution itself.
Shouldn't we distinguish between 
- a disabled button (action is not allowed)
- a busy button (action is already in progress)

WDYT?

wouter-admiraal-sonarsource:
You should ping a SC member. Anything that goes into common should be validated by both teams.

wouter-admiraal-sonarsource:
> Looks all good, but I question the solution itself.
> Shouldn't we distinguish between
> 
> * a disabled button (action is not allowed)
> * a busy button (action is already in progress)
> 
> WDYT?

Interesting thought. How would we display a busy button? With a spinner next to it? This might require a wrapper of some sorts (I'd say a `<span>`, inline-flex'd). How would this affect existing layouts? (by being inline, it shouldn't have any negative effects)

philippe-perrin-sonarsource:
> Looks all good, but I question the solution itself.
> Shouldn't we distinguish between
> 
> * a disabled button (action is not allowed)
> * a busy button (action is already in progress)
> 
> WDYT?

To be honest I thought about that but then I realized that I don't think it should be the component responsibility to render a busy state, but rather the consumer responsibility. Plus, what @wouter-admiraal-sonarsource  said. How would it affects current usages of the component ? WDYT ?

jeremy-davis-sonarsource:
Hmm... I was thinking a spinner inside of it. But you're right that the current pattern is to have it next to it. So nevermind, disabled is fine :+1:

components/controls/Toggle.tsx
```diff
@@ -36,7 +37,7 @@ export default class Toggle extends React.PureComponent<Props> {
   };
 
   handleClick = () => {
-    if (this.props.onChange) {
+    if (this.props.onChange && !this.props.disabled) {
```
siegfried-ehret-sonarsource: @philippe-perrin-sonarsource You should not need to do this.

siegfried-ehret-sonarsource: Try `<button disabled onClick={() => alert('plop')}>click me!</button>`, onClick is not called.

siegfried-ehret-sonarsource (approved):

components/controls/Toggle.tsx
philippe-perrin-sonarsource: You're damn right. Good catch.

PR #64, opened 2019-11-15
Make BoxedTabs height adjustable

wouter-admiraal-sonarsource:
See:

* https://jira.sonarsource.com/browse/SONAR-12632

Currently, the boxed tabs have a fixed height. But they should adapt based on their content.

**Checklist**

* [x] Functional validation
* [ ] ~Documentation update~
* [x] Update changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Bump to typescript@3.6.4
 - Fix BoxedTabs alignment
 - SONAR-12684 Fix glitched spinner in SearchBox
+- Don't fix BoxedTabs height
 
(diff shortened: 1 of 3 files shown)


siegfried-ehret-sonarsource (approved):

jeremy-davis-sonarsource (approved):
They're maybe a bit small now. How about 3*grid vertical padding?

wouter-admiraal-sonarsource:
You're right. Switched to `min-height` instead. I'm keeping the padding as-is; I think it looks a bit better having the same padding on top as on the sides, rather than 0 (as was the case before).

PR #66, opened 2019-11-20
SC-1379 Allo to bypass checkstatus redirection for 401 in requests

gregaubert:
**Checklist**

* [x] Write/update ITs
* [x] Functional validation
* [ ] ~Documentation update~
* [x] Update changelog

### DIFF
--- a/helpers/__tests__/request-test.ts
+++ b/helpers/__tests__/request-test.ts
@@ -21,16 +21,63 @@
 import handleRequiredAuthentication from '../handleRequiredAuthentication';
 import {
   checkStatus,
+  getJSON,
+  getText,
   parseError,
   parseJSON,
   parseText,
+  post,
+  postJSON,
   requestTryAndRepeatUntil
 } from '../request';
 
 jest.mock('../handleRequiredAuthentication', () => ({ default: jest.fn() }));
 
+const url = '/my-url';
+
 beforeEach(() => {
   jest.clearAllMocks();
+  window.fetch = jest.fn().mockResolvedValue(mockResponse({}, 200, {}));
+});
+
+describe('getJSON', () => {
+  it('should get json without parameters', async () => {
+    const response = mockResponse({}, 200, {});
+    window.fetch = jest.fn().mockResolvedValue(response);
+    getJSON(url);
+    await new Promise(setImmediate);
+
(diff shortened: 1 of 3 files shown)


helpers/request.ts
```diff
@@ -159,10 +159,10 @@ function checkApplicationVersion(response: Response): boolean {
 /**
  * Check that response status is ok
  */
-export function checkStatus(response: Response): Promise<Response> {
+export function checkStatus(response: Response, bypassRedirect = false): Promise<Response> {
```
philippe-perrin-sonarsource: For consistency with other function ?

```suggestion
export function checkStatus(response: Response, bypassRedirect?: boolean): Promise<Response> {
```

helpers/__tests__/request-test.ts
```diff
@@ -87,6 +136,45 @@ describe('parseText', () => {
...
+      expect.objectContaining({ body: 'data=test', method: 'POST' })
+    );
+  });
+});
+
+describe('post', () => {
+  it('should post without parameters and return nothing', async () => {
+    const response = new Response('{}');
+    response.json = jest.fn().mockResolvedValue({});
+    response.text = jest.fn().mockResolvedValue('');
```
philippe-perrin-sonarsource: What about a mock function that create a response based on a value, a status code, etc.

```ts
const response = mockResponse(undefined, undefined, {});
function mockResponse(headers: T.Dict<string> = {}, status = 200, value?: any): any {
  return {
    headers: { get: (prop: string) => headers[prop] },
    status,
    json: jest.fn().mockReturnValue(value ? JSON.stringify(value) : undefined),
    text: jest.fn().mockReturnValue(value ? `${value}` : undefined)
  };
}
```
Or something like that. I did not think it all the way through but a lot of code in this file is about creating the response.

philippe-perrin-sonarsource (approved):
LGTM. One comment on the test file.

helpers/request.ts
gregaubert: Yeah I wanted to make it really clear that the default value is false, but I could not do the same for the others since they have an optional parameter before this one.

gregaubert: I'll keep it consistent everywhere.

helpers/__tests__/request-test.ts
gregaubert: Good idea!

PR #67, opened 2019-11-25
Fix webapp infinite reload during BG deployement

gregaubert:
I dropped all usage of `lazyLoad` in SonarCloud to avoid infinite reload issues during Blue Green deployment. So I also dropped all it's usage in sonar-ui-common.

@wouter-admiraal-sonarsource @philippe-perrin-sonarsource @jeremy-davis-sonarsource : SQ should not be impacted by this but switching to `lazyLoadComponent` is advised, it prevent this bug from happening (which should not happen to you since you don't have BG deployement). And it better enforce props typings of the components. "Better"... I should say enforce at all, because `lazyLoad` doesn't, it uses any for the props.

**Checklist**

* [ ] ~Write/update ITs~
* [x] Functional validation
* [ ] ~Documentation update~
* [x] Update changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - SC-1402 Fix lazyLoadComponent error handling
 - SC-1402 requestTryAndRepeatUntil now returns the error response after the max number of tries
+- SC-1402 Replace usage of lazyLoad by lazyLoadComponent in Select
 
 ## 0.0.39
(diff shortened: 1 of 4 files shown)


siegfried-ehret-sonarsource (approved):

wouter-admiraal-sonarsource (approved):
LGTM!

PR #70, opened 2019-12-03
SONAR-12717 Add a helper to hide footer

jeremy-davis-sonarsource:
For the new Hotspots page that has no footer

**Checklist**

* [x] Functional validation
* [x] Update changelog

### DIFF
--- a/components/controls/BoxedTabs.tsx
+++ b/components/controls/BoxedTabs.tsx
@@ -26,11 +26,17 @@ export interface BoxedTabsProps<K> {
   tabs: Array<{ key: K; label: React.ReactNode }>;
 }
 
-const baseBorder = ({ theme }: ThemedProps) => `1px solid ${theme.colors.barBorderColor}`;
-
 const TabContainer = styled.div`
   display: flex;
   flex-direction: row;
+`;
+
+const baseBorder = ({ theme }: ThemedProps) => `1px solid ${theme.colors.barBorderColor}`;
+
+const highlightHoverMixin = ({ theme }: ThemedProps) => `
+  &:hover {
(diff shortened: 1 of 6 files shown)


wouter-admiraal-sonarsource (approved):
LGTM, but _why_ are we removing the footer on specific pages? Especially on a single page? Did Matt intend that?

components/controls/BoxedTabs.tsx
```diff
@@ -43,13 +48,11 @@ const StyledTab = styled.button<{ active: boolean }>`
...
   min-height: 56px;
-  cursor: pointer;
+  ${props => !props.active && 'cursor: pointer;'}
   outline: 0;
   padding: calc(2 * ${props => props.theme.sizes.gridSize});
 
-  &:hover {
-    background-color: ${props => props.theme.colors.barBackgroundColorHighlight};
-  }
+  ${props => (!props.active ? highlightHoverMixin(props.theme) : null)}
```
philippe-perrin-sonarsource: I think you can avoid passing the `props.theme` and assume in the mixin that the parameters is `props: { theme: ThemedProps }`.

philippe-perrin-sonarsource: NP: I think you can avoid passing the `props.theme` and assume in the mixin that the parameter is `props: { theme: ThemedProps }`.

philippe-perrin-sonarsource (approved):
LGTM. NP only.

gregaubert:
@jeremy-davis-sonarsource Could you also share the related code on SQ side about the helper to hide a footer ? Because this ends up in sonar-ui-common but we could not use it in SC currently since the class is defined on SQ side only.

PR #71, opened 2019-12-19
SONAR-12637 Allow ActionsDropdown to be positioned

wouter-admiraal-sonarsource:
Allow the `ActionsDropdown` component to be positioned via the `overlayPlacement` prop, just as for the `Dropdown` component.

**Checklist**

* [x] Functional validation
* [ ] ~Documentation update~
* [x] Update changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- SONAR-12637 Allow ActionsDropdown to be positioned like any other dropdown
+
 ## 0.0.41
 
 - SONAR-12717 Add a helper to hide footer
--- a/components/controls/__tests__/ActionsDropdown-test.tsx
+++ b/components/controls/__tests__/ActionsDropdown-test.tsx
@@ -1,0 +1,79 @@
+/*
+ * Sonar UI Common
+ * Copyright (C) 2019-2019 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+/* eslint-disable sonarjs/no-duplicate-string */
+import { shallow } from 'enzyme';
+import * as React from 'react';
+import { click } from '../../../helpers/testUtils';
+import { PopupPlacement } from '../../ui/popups';
+import ActionsDropdown, {
+  ActionsDropdownDivider,
+  ActionsDropdownItem,
+  ActionsDropdownProps
+} from '../ActionsDropdown';
+
+describe('ActionsDropdown', () => {
+  it('should render correctly', () => {
+    expect(shallowRender()).toMatchSnapshot();
+    expect(shallowRender({ small: false })).toMatchSnapshot();
+  });
+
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - SC-1402 Replace usage of lazyLoad by lazyLoadComponent in Select
 - SONAR-12679 Introduce new BranchIcon
 - SONAR-12633 New BackIcon
+- SONAR-12637 Allow ActionsDropdown to be positioned like any other dropdown
```
gregaubert: You might not be up to date with master ?
We are at version 41 I think

components/controls/__tests__/ActionsDropdown-test.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { click } from '../../../helpers/testUtils';
+import { PopupPlacement } from '../../ui/popups';
+import ActionsDropdown, { ActionsDropdownDivider, ActionsDropdownItem } from '../ActionsDropdown';
+
+describe('ActionsDropdown', () => {
+  it('should render correctly', () => {
+    expect(
+      shallow(
+        <ActionsDropdown
+          // eslint-disable-next-line react/no-children-prop
```
gregaubert: Is it raising blocking issues ? Quality gate wise ? If not maybe it's better to just have them reported on SonarQube instead of polluting the code with those comments ?

gregaubert (approved):
Looks good! Just wondering about those eslint comments...

components/controls/__tests__/ActionsDropdown-test.tsx
jeremy-davis-sonarsource: Better yet, pass the children as actual children!

jeremy-davis-sonarsource (approved):
LGTM!

components/controls/__tests__/ActionsDropdown-test.tsx
gregaubert: Ah yeah indeed xD

components/controls/__tests__/ActionsDropdown-test.tsx
```diff
@@ -0,0 +1,76 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+/* eslint-disable sonarjs/no-duplicate-string */
```
gregaubert, line 20: My previous comment still hold for this one, is it really necessary ?

CHANGELOG.md
wouter-admiraal-sonarsource: Ah, thought I had pulled. Will do.

components/controls/__tests__/ActionsDropdown-test.tsx
wouter-admiraal-sonarsource: Yeah, leftover from a quick refactor. It was part of a `shallowRender()` call, and I wanted to let the children be passed as props. But not relevant here. Will change.

components/controls/__tests__/ActionsDropdown-test.tsx
wouter-admiraal-sonarsource: Yes, it will raise a Code Smell on "should render correctly" being duplicated 3 times. I guess we can close them manually in SQ, but I find that, in tests, we very often get these (like for using `"foo"` or a date `"2019-10-02"`). So sometimes, I just add this comment. It prevents pollution in VSCode, and doesn't raise Code Smells in SQ.

wouter-admiraal-sonarsource: BTW, I looked at disabling ESLint rules for certain file patterns, but couldn't find a way. Duplicated strings are very common in our tests, and I'm not in favor of having something like `const foo = "foo";` at the top of a test, just so we don't duplicate that string in certain assertions. I prefer being explicit. But again, VSCode shows warnings, so it's kind of annoying... I personally don't mind such comments, when they're there for a good reason.

gregaubert: Yeah I'm not in favor of having `const foo = "foo";` either. I guess eslint comments on top of the files are ok, I just really don't like them in the middle of the code.

PR #72, opened 2019-12-20
Abort release process if publishing fails

wouter-admiraal-sonarsource:
Right now, if the publishing to NPM fails due to an incorrect password, the release process won't interrupt, and we push the changes to Git anyway. We should abort, and let the person doing the release clean stuff up and fix the problem.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -25,7 +25,20 @@ git add CHANGELOG.md package.json
 git commit -m "Prepare version $1"
 git tag $1
 yarn publish ./build/dist/sonar-ui-common-v$1.tgz
-npmrc default
+
+if [ $? -gt 0 ]
+  then
+    echo "Publish failed, aborting"
+    git tag -d $1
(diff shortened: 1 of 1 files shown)


scripts/release.sh
```diff
@@ -25,6 +25,14 @@ git add CHANGELOG.md package.json
 git commit -m "Prepare version $1"
 git tag $1
 yarn publish ./build/dist/sonar-ui-common-v$1.tgz
+
+if [ $? -gt 0 ]
+  then
+    echo "Publish failed, aborting"
+    npmrc default
```
gregaubert, line 35: Should we also revert the commit and tag ?

gregaubert (approved):
Nice improvement! Thanks!

scripts/release.sh
wouter-admiraal-sonarsource: Good idea! I see now that it really messes stuff up if you run it again afterwards 😑 .

PR #73, opened 2019-12-23
a11y hardening

wouter-admiraal-sonarsource:
Improve accessibility on some components we have.

**Checklist**

* [x] Functional validation
* [ ] ~Documentation update~
* [x] Update changelog

### DIFF
--- a/components/controls/buttons.tsx
+++ b/components/controls/buttons.tsx
@@ -99,6 +99,8 @@ export function ResetButtonLink(props: T.Omit<ButtonProps, 'type'>) {
 }
 
 export interface ButtonIconProps extends ButtonProps {
+  'aria-label'?: string;
+  'aria-labelledby'?: string;
   className?: string;
   color?: string;
   onClick?: () => void;
--- a/components/ui/Level.tsx
+++ b/components/ui/Level.tsx
@@ -22,19 +22,28 @@ import * as React from 'react';
 import { formatMeasure } from '../../helpers/measures';
 import './Level.css';
 
-interface Props {
+export interface LevelProps {
+  'aria-label'?: string;
+  'aria-labelledby'?: string;
   className?: string;
   level: string;
   small?: boolean;
   muted?: boolean;
 }
 
-export default function Level(props: Props) {
+export default function Level(props: LevelProps) {
   const formatted = formatMeasure(props.level, 'LEVEL');
   const className = classNames(props.className, 'level', 'level-' + props.level, {
     'level-small': props.small,
     'level-muted': props.muted
   });
 
-  return <span className={className}>{formatted}</span>;
+  return (
+    <span
+      aria-label={props['aria-label']}
+      aria-labelledby={props['aria-labelledby']}
+      className={className}>
+      {formatted}
+    </span>
+  );
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Allow icon buttons to have an ARIA label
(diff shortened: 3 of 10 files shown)


components/controls/buttons.tsx
```diff
@@ -99,6 +99,7 @@ export function ResetButtonLink(props: T.Omit<ButtonProps, 'type'>) {
 }
 
 export interface ButtonIconProps extends ButtonProps {
+  'aria-label'?: string;
```
siegfried-ehret-sonarsource, line 102: How about an `aria-labelledby` also ?

siegfried-ehret-sonarsource (approved):

components/ui/Level.tsx
```diff
@@ -22,19 +22,24 @@ import * as React from 'react';
 import { formatMeasure } from '../../helpers/measures';
 import './Level.css';
 
-interface Props {
+export interface LevelProps {
+  'aria-label'?: string;
```
philippe-perrin-sonarsource, line 26: NP: I get that the idea is to write an `aria-label` the same way on component & classic html tag, but don't you think it might be confusing ? The component itself won't have an `aria-*' attribute, some node inside it will. IMO, it makes it more difficult to know if it is a react component's props or an html attribute.

philippe-perrin-sonarsource: Interesting read: https://github.com/airbnb/javascript/tree/master/react#naming

philippe-perrin-sonarsource (approved):
LGTM

components/ui/Level.tsx
wouter-admiraal-sonarsource: I think we have 1 or 2 components that have something like `ariaLabel` as a prop. And I didn't even realize we _could_ name props this way. However, I noticed it was used by react virtualized, and I liked it.

React did not change the way we write `aria-*` attributes, so you still use `aria-label`, `aria-hidden` on components, and they do what you expect... The only drawback I see with my approach here, is that you could expect to have access to _any_ `aria-*` attribute, where in fact I whitelist them. But then, you probably wouldn't want to use the others anyway in this context.

I don't really know. I see your point, but I'm not sure I like `ariaLabel` for this purpose, as we would start having mixed `aria-label` and `ariaLabel` props, depending on whether we use HTML elements or React Components... I'm more in favor of keeping this consistent.
 @siegfried-ehret-sonarsource @jeremy-davis-sonarsource quick thoughts?

siegfried-ehret-sonarsource: I have absolutely no idea ^^'.

wouter-admiraal-sonarsource: Discussed with Jay: React [defines them using strings](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L1502), and having camelCase or pipe-case depending on the context seems like a bad idea (especially since certain custom components _can_ have `aria-label` because of the usage of `{...props}`, as demonstrated by the [`FavoriteButton`](https://github.com/SonarSource/sonar-ui-common/blob/master/components/controls/FavoriteButton.tsx#L45) component 🙀 ).

So, `'aria-label'` it is.

philippe-perrin-sonarsource: Great. If you think this worth being documented (is there a rule we can come up with ?), feel free to add this subject to our ongoing technical discussion/documentation.

wouter-admiraal-sonarsource: I'm actually wondering if we could improve on this with a custom HOC 🤔 . Which would allow us to use _all_ `aria-*` attributes. That could be even better.

PR #74, opened 2019-12-27
Prevent click event from bubbling when using ActionsDropdown

wouter-admiraal-sonarsource:
A click inside a `ConfirmModal` on one of the buttons (submit or cancel) will bubble up to the parent component (thanks to React's event handling, because the modal is not even an actual child node in the DOM).

This is:

- not always desirable (on the Activity page, editing an event and saving/canceling triggers a click on the analysis... this is not the case if you simply hit Esc)
- probably never even useful

It should remain contained within the modal.

click_bubbles_through_modal

**Checklist**

* [x] Functional validation
* [ ] ~Documentation update~
* [x] Update changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - BoxedTabs : add support for external css class
+- ConfirmModal: prevent click events from bubbling to the parent component
 
 ## 0.0.43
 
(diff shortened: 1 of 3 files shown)


siegfried-ehret-sonarsource (approved):

wouter-admiraal-sonarsource:
Ouch, it's even worse than I thought 😅 . Even clicking on a field inside the modal triggers the click event... fixup commit uses `ClickEventBoundary`.

philippe-perrin-sonarsource (approved):
LGTM. Be careful with the fixup commit

PR #79, opened 2020-01-08
SONAR-12633 Add new EscKeydownHandler component

wouter-admiraal-sonarsource:
Add a new handler, similar to `OutsideClickHandler` and `DocumentClickHandler`, in order to listen to a ESCAPE keydown event.

**Checklist**

* [x] Functional validation
* [ ] ~Documentation update~
* [x] Update changelog

### DIFF
--- a/components/controls/EscKeydownHandler.tsx
+++ b/components/controls/EscKeydownHandler.tsx
@@ -17,32 +17,32 @@
  * along with this program; if not, write to the Free Software Foundation,
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
-import { shallow } from 'enzyme';
 import * as React from 'react';
-import Toggler from '../Toggler';
+import { KeyCodes } from '../../helpers/keycodes';
 
-it('should render only children', () => {
-  expect(shallowRender({ open: false })).toMatchSnapshot();
-});
+interface Props {
+  children: React.ReactNode;
+  onKeydown: () => void;
+}
 
-it('should render children and overlay', () => {
-  expect(shallowRender()).toMatchSnapshot();
-});
+export default class EscKeydownHandler extends React.Component<Props> {
+  componentDidMount() {
+    setTimeout(() => {
+      document.addEventListener('keydown', this.handleKeyDown, false);
+    }, 0);
+  }
 
-it('should render when closeOnClick=true', () => {
-  expect(shallowRender({ closeOnClick: true })).toMatchSnapshot();
-});
+  componentWillUnmount() {
+    document.removeEventListener('keydown', this.handleKeyDown, false);
+  }
 
-it('should not render click wrappers', () => {
-  expect(
-    shallowRender({ closeOnClick: false, closeOnClickOutside: false, closeOnEscape: false })
-  ).toMatchSnapshot();
-});
+  handleKeyDown = (event: KeyboardEvent) => {
+    if (event.keyCode === KeyCodes.Escape) {
+      this.props.onKeydown();
+    }
+  };
 
-function shallowRender(props?: Partial<Toggler['props']>) {
-  return shallow(
-    <Toggler onRequestClose={jest.fn()} open={true} overlay={<div id="overlay" />} {...props}>
-      <div id="toggle" />
-    </Toggler>
-  );
+  render() {
+    return this.props.children;
+  }
 }
--- a/components/controls/__tests__/EscKeydownHandler-test.tsx
+++ b/components/controls/__tests__/EscKeydownHandler-test.tsx
@@ -17,32 +17,31 @@
  * along with this program; if not, write to the Free Software Foundation,
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
+
 import { shallow } from 'enzyme';
 import * as React from 'react';
-import Toggler from '../Toggler';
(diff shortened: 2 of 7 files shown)


components/controls/EscKeydownHandler.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { KeyCodes } from '../../helpers/keycodes';
+
+interface Props {
+  children: React.ReactNode;
+  onKeydown: () => void;
+}
+
+export default class EscKeydownHandler extends React.Component<Props> {
+  componentDidMount() {
+    setTimeout(() => {
```
gregaubert, line 30: Why do we need this timeout at 0 ?

components/controls/EscKeydownHandler.tsx
```diff
@@ -0,0 +1,48 @@
...
+      document.addEventListener('keydown', this.handleKeyDown, false);
+    }, 0);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keydown', this.handleKeyDown, false);
+  }
+
+  handleKeyDown = (event: KeyboardEvent) => {
+    if (event.keyCode === KeyCodes.Escape) {
```
gregaubert, line 40: `keyCode` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode), seems like the replacement is `key` or `code`. Maybe other keydown handler should be updated too ?

gregaubert (approved):
Looking good! I'm just wondering about the deprecated `keyCode`...

components/controls/EscKeydownHandler.tsx
wouter-admiraal-sonarsource: Not sure, it was the case in `Toggler`, where this code basically comes from, so I thought I'd keep it, in case it was for some edge case somewhere.

components/controls/EscKeydownHandler.tsx
wouter-admiraal-sonarsource: Yeah, went for the quick'n'easy way. The thing is, `code` isn't supported in IE11, and `key` is different ("Esc" rather than "Escape"). You think we should still use the new API? I can check for both cases.

gregaubert: And not supported in Edge either no ?
I think we should use the new API sooner or later because we might end up with a recent browser dropping it, but I don't see any urgency to do it and indeed it does seem more complicated to use the new API. Up to you, I'm fine in both cases.

components/controls/EscKeydownHandler.tsx
gregaubert: Ok that's weird we don't do such stuff in the other places where we bind key events... And I guess we lost the history about this when moving to sonar-ui-common...

components/controls/EscKeydownHandler.tsx
philippe-perrin-sonarsource: FYI: we use `keyCode` in many places, but through the `React.KeyboardEvent` interface, where it is not deprecated. Here's your solution 🤣

components/controls/EscKeydownHandler.tsx
philippe-perrin-sonarsource: I believe this is a common trick to run code on another thread. But I don't understand why it is used here.

components/controls/EscKeydownHandler.tsx
wouter-admiraal-sonarsource: I'll keep it like this for now. And as we're using the `KeyCodes` enum, it will be easy to find this occurrence.

components/controls/EscKeydownHandler.tsx
wouter-admiraal-sonarsource: Yeah... And this has an impact on many places (because of `Toggler`). So, I'm not confident  changing this is safe 😅

components/controls/__tests__/EscKeydownHandler-test.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+it('should render correctly', () => {
+  expect(shallowRender()).toMatchSnapshot();
+});
+
+it('should correctly trigger the keydown handler when hitting Esc', () => {
+  const onKeydown = jest.fn();
+  shallowRender({ onKeydown });
+  jest.runAllTimers();
+  keydown('esc');
```
philippe-perrin-sonarsource: keydown(KeyCodes.Escape);

helpers/testUtils.ts
```diff
@@ -71,6 +71,7 @@ export function change(element: ShallowWrapper | ReactWrapper, value: string, ev
 
 export const KEYCODE_MAP: { [keycode: number]: string } = {
   13: 'enter',
+  27: 'esc',
```
philippe-perrin-sonarsource: Removed in favor of the enum value.

philippe-perrin-sonarsource (approved):
LGTM. Some comments

wouter-admiraal-sonarsource:
@gregaubert Philippe's comment is right, we should at least use `KeyCodes` with `keydown()`, which is just as readable as the string aliases ('up', 'down', etc).

I suggest adding a warning that we deprecate using things like `keydown('up');`. See 2nd commit. WDYT?

gregaubert:
> I suggest adding a warning that we deprecate using things like `keydown('up');`. See 2nd commit. WDYT?

Good idea!

PR #81, opened 2020-01-09
SONAR-12860: Add custom type for RadioToggle and size glitch

mathieu-suen-sonarsource:
**Checklist**
* [x] Functional validation
* [ ] Public documentation update
* [x] Ping a member of the other team (SQ/SC)

### DIFF
--- a/components/controls/RadioToggle.css
+++ b/components/controls/RadioToggle.css
@@ -53,6 +53,7 @@
   height: calc(var(--controlHeight) - 2px);
   line-height: calc(var(--controlHeight) - 2px);
   cursor: pointer;
+  font-weight: 600;
 }
 
 .radio-toggle input[type='radio'] {
@@ -62,7 +63,6 @@
 .radio-toggle input[type='radio']:checked + label {
   background-color: var(--darkBlue);
   color: #fff;
-  font-weight: 600;
 }
 
 .radio-toggle input[type='radio']:disabled + label {
--- a/components/controls/RadioToggle.tsx
+++ b/components/controls/RadioToggle.tsx
@@ -22,19 +22,20 @@ import * as React from 'react';
 import './RadioToggle.css';
 import Tooltip from './Tooltip';
 
+type ToggleValueType = string | number | boolean;
 interface Option {
   disabled?: boolean;
   label: string;
   tooltip?: string;
-  value: string;
+  value: ToggleValueType;
 }
 
 interface Props {
   className?: string;
   name: string;
-  onCheck: (value: string) => void;
+  onCheck: (value: ToggleValueType) => void;
   options: Option[];
-  value?: string;
+  value?: ToggleValueType;
 }
 
 export default class RadioToggle extends React.PureComponent<Props> {
@@ -43,24 +44,18 @@ export default class RadioToggle extends React.PureComponent<Props> {
     value: null
   };
 
-  handleChange = (e: React.SyntheticEvent<HTMLInputElement>) => {
-    const newValue = e.currentTarget.value;
-    this.props.onCheck(newValue);
-  };
-
   renderOption = (option: Option) => {
     const checked = option.value === this.props.value;
     const htmlId = this.props.name + '__' + option.value;
     return (
-      <li key={option.value}>
+      <li key={option.value.toString()}>
         <input
           checked={checked}
           disabled={option.disabled}
           id={htmlId}
           name={this.props.name}
-          onChange={this.handleChange}
+          onChange={() => this.props.onCheck(option.value)}
           type="radio"
-          value={option.value}
         />
         <Tooltip overlay={option.tooltip || undefined}>
           <label htmlFor={htmlId}>{option.label}</label>
--- a/components/controls/__tests__/RadioToggle-test.tsx
+++ b/components/controls/__tests__/RadioToggle-test.tsx
@@ -23,33 +23,72 @@ import { change } from '../../../helpers/testUtils';
 import RadioToggle from '../RadioToggle';
 
 it('renders', () => {
-  expect(shallow(getSample())).toMatchSnapshot();
+  expect(shallowRender()).toMatchSnapshot();
 });
 
 it('calls onCheck', () => {
   const onCheck = jest.fn();
-  const wrapper = shallow(getSample({ onCheck }));
-  change(wrapper.find('input[value="two"]'), 'two');
+  const wrapper = shallowRender({ onCheck });
+  change(wrapper.find('input[id="sample__two"]'), '');
   expect(onCheck).toBeCalledWith('two');
+});
+
+it('handles numeric values', () => {
+  const onCheck = jest.fn();
+  const wrapper = shallowRender({
+    onCheck,
+    options: [
+      { value: 1, label: 'first', tooltip: 'foo' },
+      { value: 2, label: 'second', tooltip: 'bar' }
+    ],
+    value: 1
+  });
+  change(wrapper.find('input[id="sample__2"]'), '');
+  expect(onCheck).toBeCalledWith(2);
+});
+
+it('handles boolean values', () => {
+  const onCheck = jest.fn();
+  const wrapper = shallowRender({
+    onCheck,
+    options: [
+      { value: true, label: 'yes', tooltip: 'foo' },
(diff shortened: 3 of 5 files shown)


components/controls/__tests__/RadioToggle-test.tsx
```diff
@@ -29,10 +29,25 @@ it('renders', () => {
 it('calls onCheck', () => {
   const onCheck = jest.fn();
   const wrapper = shallow(getSample({ onCheck }));
-  change(wrapper.find('input[value="two"]'), 'two');
+  change(wrapper.find('input[value="second"]'), 'second');
   expect(onCheck).toBeCalledWith('two');
 });
 
+it('calls with number type', () => {
```
jeremy-davis-sonarsource: `calls with number type` doesn't mean anything :smile: 
How about `handles numeric values`.
Also you should check that setting the value selects the expected option

components/controls/RadioToggle.tsx
```diff
@@ -18,52 +18,58 @@
...
+export default class RadioToggle<T = string> extends React.PureComponent<Props<T>> {
   static defaultProps = {
     disabled: false,
     value: null
   };
 
   handleChange = (e: React.SyntheticEvent<HTMLInputElement>) => {
     const newValue = e.currentTarget.value;
-    this.props.onCheck(newValue);
+    // Let assum that if we can't found the option we have unselect
```
jeremy-davis-sonarsource: `We assume that if we can't find the option, it was unselected`.
Although I'm not sure this is an actual use-case. I would call `onCheck` only if the option is found.
Let's see what Greg thinks.

jeremy-davis-sonarsource (approved):
Nice work! Fix the code smell as well as the test, and you'll be good.
Let's also see what Greg says before running with it

components/controls/RadioToggle.css
```diff
@@ -62,7 +62,6 @@
 .radio-toggle input[type='radio']:checked + label {
   background-color: var(--darkBlue);
   color: #fff;
-  font-weight: 600;
```
gregaubert, line 65: I found the active state really harder to read without the bold, are we sure we want to remove it ?
On the other way I agree that it's ugly that the full radio toggle change size because of the bold.

components/controls/RadioToggle.tsx
```diff
@@ -18,52 +18,58 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 import * as classNames from 'classnames';
+import * as _ from 'lodash';
```
gregaubert: Same comment as the previous PR, import only what you need from lodash

components/controls/RadioToggle.tsx
```diff
@@ -18,52 +18,58 @@
...
 }
 
-interface Props {
+interface Props<T = string> {
   className?: string;
   name: string;
-  onCheck: (value: string) => void;
-  options: Option[];
-  value?: string;
+  onCheck: (value: T | undefined) => void;
```
gregaubert: I don't think we should have `undefined` possible there. If `onCheck` is called that means a option has been selected...
I don't think there currently is a use case where we want to un-check an option.

components/controls/RadioToggle.tsx
gregaubert: I think we should not support this use case, `handleChange` can only be called from an input, so there should always be a correct value.
I think we should call the `onCheck` only if the option is found.

components/controls/RadioToggle.tsx
```diff
@@ -18,52 +18,58 @@
...
     disabled: false,
     value: null
   };
 
   handleChange = (e: React.SyntheticEvent<HTMLInputElement>) => {
     const newValue = e.currentTarget.value;
-    this.props.onCheck(newValue);
+    // Let assum that if we can't found the option we have unselect
+    // RadioToggle
+    const selectedOption = _.find(this.props.options, { label: newValue });
```
gregaubert: I think using [`Array.find`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) is better, it's as easy to read.

Except for the native `sort` that is soooo baaaad, I think it's always better to use a native method than one from lodash if it does the same thing.
I guess it can only be faster than the lodash one and it reduces the amount of code that needs to be put in chunks.

components/controls/RadioToggle.tsx
```diff
@@ -18,52 +18,58 @@
...
           id={htmlId}
           name={this.props.name}
           onChange={this.handleChange}
           type="radio"
-          value={option.value}
+          value={option.label}
         />
         <Tooltip overlay={option.tooltip || undefined}>
-          <label htmlFor={htmlId}>{option.label}</label>
+          <label title={option.label} htmlFor={htmlId}>
```
gregaubert: Should we really have both a tooltip and title ? I'd say in most cases we don't want a title on those `RadioToggle`.
(I'm thinking about how we use them in SC)

components/controls/RadioToggle.tsx
```diff
@@ -18,52 +18,58 @@
...
+    this.props.onCheck(selectedOption && selectedOption.value);
   };
 
-  renderOption = (option: Option) => {
+  renderOption = (option: Option<T>) => {
     const checked = option.value === this.props.value;
     const htmlId = this.props.name + '__' + option.value;
     return (
-      <li key={option.value}>
+      <li key={option.label}>
```
gregaubert: I'm not sure we should really use the label for this and the value... I guess it's only a coincidence that the label is a `string` here, it could be as well be a `ReactNode` to let some flexibility on what we can display...

In fact, reading the ticket, I'm not sure to understand why we want to be able to handle anything, is there a use case behind it @jeremy-davis-sonarsource ? Maybe it should be limited to string, number or an object that would have a key. So that we can use the key of that object as the value.

components/controls/__tests__/RadioToggle-test.tsx
```diff
@@ -29,10 +29,25 @@ it('renders', () => {
...
+  const onCheck = jest.fn();
+  const wrapper = shallow(
+    getSample({
```
gregaubert: Since you are working on this, could you rename this to `shallowRender` and type correctly the props ? :smile:

gregaubert (changes requested):
I'm not sure about the `label` thing and maybe we should re-discuss the api we want for this component. 
I currently don't see the use case behind supporting any type ? Also we only have tests for number and string currently, we should also test more complex object if we really want to support them.

components/controls/RadioToggle.tsx
philippe-perrin-sonarsource: I'm also wondering why we need to improve the typing of this value. I knew it but now I can't remember. Can you enlighten us ?

components/controls/RadioToggle.tsx
philippe-perrin-sonarsource: Indeed it looks like this is too much. Didn't check the result though.

components/controls/RadioToggle.css
philippe-perrin-sonarsource: Obviously, I personally think it's ok. You should delete the task SONAR-12906 if SONAR-12860 covers both issues.

gregaubert: Maybe check with Matt ?

mathieu-suen-sonarsource: What if we always keep bold ?

components/controls/RadioToggle.tsx
mathieu-suen-sonarsource: As we have discuss we can use controllled componant.

components/controls/RadioToggle.css
gregaubert: yeah why not

gregaubert (approved):
Looks good!
You might just want to split your two ticket changes in two commits, I don't know if it's still the case or not but as I remember you guys have a tools to generate release notes from commits and I don't know if it can understand multiple ticket number per message, or a ticket number in the middle of a message.
Anyway it's cleaner to have commits per tickets and the related changes in each commits.

mathieu-suen-sonarsource:
@gregaubert good point. I did the split.

PR #83, opened 2020-01-13
Clean button-group css

gregaubert:
While doing some cleaning of `isSonarCloud` in SC repo I found out that all this css about `button-group` is SQ only. And I think it is kind of replaced by the RadioToggle in fact, I think it should be moved over to SQ repo.

**Checklist**

* [ ] ~Functional validation~
* [ ] ~Documentation update~
* [x] Update changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-
+- Clean button-group css
 
 ## 0.0.47
 
--- a/components/controls/buttons.css
+++ b/components/controls/buttons.css
@@ -184,57 +184,6 @@
   transform: translateY(-2px);
 }
 
-/* #region .button-group */
-.button-group {
-  display: inline-block;
-  vertical-align: middle;
-  font-size: 0;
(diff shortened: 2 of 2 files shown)


jeremy-davis-sonarsource:
I'm not sure we really need to exclude things that aren't shared from sonar-ui-common.
That said, it might be interesting to get rid of this class completely and create a component (or use RadioToggle, if applicable)... In any case, I think the change/fix/cleanup should come from a SonarQube effort.

gregaubert:
> I'm not sure we really need to exclude things that aren't shared from sonar-ui-common.

Indeed, if things are clean and generic enough.

> That said, it might be interesting to get rid of this class completely and create a component (or use RadioToggle, if applicable)... In any case, I think the change/fix/cleanup should come from a SonarQube effort.

Yeah the initiative should come from SQ side indeed, I'll close this PR and just add a todo comment saying that usage should be dropped. Now I just need to keep an eye open so that no-one tries to use this on our side! :eyes:

PR #84, opened 2020-01-14
Scroll chaining

wouter-admiraal-sonarsource:
With [SONAR-12452](https://jira.sonarsource.com/browse/SONAR-12452), we have an issue where several scrolls can be triggered at once. Browsers don't support this well, which breaks the flow in the UI.

This turns scroll requests into a queue, which gets immediately processed when called, but prevents multiple scrolls from overlapping. 

**Related PR:**

* https://github.com/SonarSource/sonar-enterprise/pull/2421

**Checklist**

* [x] Functional validation
* [ ] ~Documentation update~
* [x] Update changelog

### DIFF
--- a/helpers/__tests__/scrolling-test.ts
+++ b/helpers/__tests__/scrolling-test.ts
@@ -183,12 +183,37 @@ describe('scrollHorizontally', () => {
   });
 });
 
-const mockGetBoundingClientRect = (overrides: Partial<ClientRect>) => () => ({
-  bottom: 0,
-  height: 0,
-  left: 0,
-  right: 0,
-  top: 0,
-  width: 0,
-  ...overrides
+it('correctly queues and processes multiple scroll calls', async () => {
+  const element1 = document.createElement('a');
+  const element2 = document.createElement('a');
+  document.body.appendChild(element1);
+  document.body.appendChild(element2);
+  element1.getBoundingClientRect = mockGetBoundingClientRect({ left: 840, right: 845 });
+  element2.getBoundingClientRect = mockGetBoundingClientRect({ top: -10, bottom: 10 });
+
+  window.scrollTo = jest.fn();
+
+  scrollHorizontally(element1, {});
+  scrollToElement(element2, { smooth: false });
+
+  jest.runAllTimers();
+  await Promise.resolve(setImmediate);
+  await Promise.resolve(setImmediate);
+
+  expect(window.scrollTo).toBeCalledTimes(11);
+
+  scrollHorizontally(element1, {});
+  jest.runAllTimers();
+  expect(window.scrollTo).toBeCalledTimes(21);
 });
+
+const mockGetBoundingClientRect = (overrides: Partial<ClientRect>) => () =>
+  ({
+    bottom: 0,
+    height: 0,
+    left: 0,
+    right: 0,
+    top: 0,
+    width: 0,
+    ...overrides
+  } as DOMRect);
--- a/helpers/scrolling.ts
+++ b/helpers/scrolling.ts
@@ -17,7 +17,6 @@
  * along with this program; if not, write to the Free Software Foundation,
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
-import { debounce } from 'lodash';
 
 const SCROLLING_DURATION = 100;
 const SCROLLING_INTERVAL = 10;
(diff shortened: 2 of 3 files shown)


wouter-admiraal-sonarsource:
Asking for 4 reviewers, as this is an important change 😉

helpers/__tests__/scrolling-test.ts
```diff
@@ -183,12 +183,38 @@ describe('scrollHorizontally', () => {
...
+  element2.getBoundingClientRect = mockGetBoundingClientRect({ top: -10, bottom: 10 });
+
+  window.scrollTo = jest.fn();
+
+  scrollHorizontally(element1, {});
+  scrollToElement(element2, { smooth: false });
+
+  jest.runAllTimers();
+  await Promise.resolve(setImmediate);
+  jest.runAllTimers();
```
jeremy-davis-sonarsource: NP: this can be done once, it runs the recursive timers

jeremy-davis-sonarsource (approved):
LGTM
I was thinking about handling a kind of debounce/cancel (like using a map to keep track of target elements) but it's probably overkill

helpers/scrolling.ts
```diff
@@ -139,18 +145,52 @@ export function scrollHorizontally(
...
     } else {
-      scrollElement(parent, goal, scroll.y);
+      addToScrollQueue(scrollElement, parent, goal, scroll.y);
     }
   }
 }
+
+type ScrollFunction = (element: Element | Window, x: number, y?: number) => Promise<void>;
+
+type ScrollQueueItem = {
```
gregaubert: How do you choose between a type and an interface ? Why type instead of interface here for exemple ?

helpers/scrolling.ts
```diff
@@ -139,18 +145,52 @@ export function scrollHorizontally(
...
+  queue.push({ fn, element, x, y });
+  if (!queueRunning) {
+    processQueue();
+  }
+}
+
+function processQueue() {
+  if (queue.length > 0) {
+    queueRunning = true;
+    const { fn, element, x, y } = queue.splice(0, 1)[0];
```
gregaubert: Maybe use `shift` here instead of `splice`, because that's in fact what you are doing.

helpers/scrolling.ts
```diff
@@ -139,18 +145,52 @@ export function scrollHorizontally(
...
+    fn(element, x, y).finally(processQueue);
```
gregaubert: Do we really support `finally` ? Did you add polyfill for it in SQ build ?

gregaubert (commented):
Seems good! Just a few minor comments, and I might need to update our build for the use of `finally`...
Just one last question, how and where in SC can I reproduce the bugs it's supposed to fix ?

helpers/scrolling.ts
wouter-admiraal-sonarsource: Ha! Good question :-). I don't know. I guess I started with something much simpler, and just added to it, keeping `type`. I don't have any rule for choosing one or the other. I guess `interface` is just as appropriate here.

Anybody else have thoughts about this?

helpers/scrolling.ts
wouter-admiraal-sonarsource: Of course 👍 !

helpers/scrolling.ts
wouter-admiraal-sonarsource: We do, we added a polyfill for it. However, now I'm wondering if we need the polyfill for the SUC build as well 🤔 ... We shouldn't, right?

wouter-admiraal-sonarsource:
> @gregaubert  Just one last question, how and where in SC can I reproduce the bugs it's supposed to fix ?

You'll need a multi-location issue... We have a few in SQ, but mainly because of the security engine. It will be difficult, I guess, to find something similar in SC. A multi-location bug might not have enough locations to really see the problem in action.

helpers/scrolling.ts
philippe-perrin-sonarsource: There's some difference described [here](https://medium.com/@martin_hotell/interface-vs-type-alias-in-typescript-2-7-2a8f1777af4c). Although it's an interesting reading, I agree we should use interface unless we truly have to use something else.

helpers/scrolling.ts
philippe-perrin-sonarsource: Do we add something for it ? I don't personally. But I checked and it worked on IE & EDGE.

gregaubert: Ok thanks, I'll add a polyfill for it in SC too... Even with the drop of legacy browsers it seems we still support browsers that don't have `finally`.

I guess as soon as we want to make it public we will need to have a cleaner build indeed, but for now we don't have any polyfill at all and the build is very basic (just doing a `tsc`). So it rely on the fact that SQ and SC have the polyfill...

gregaubert (approved):
Tested on SC and it works very well! Well done!

philippe-perrin-sonarsource (approved):
LGTM

helpers/scrolling.ts
gregaubert: I've been investigating that polyfill thing and the current situation is a bit weird.
For SQ the modern build has no polyfill at all and would not need any for `finally` since it's supposed to support the last 2 browsers versions.
The legacy build has polyfill and support IE11 and last 3 browsers version.
BUT the switch between legacy and modern build is done based on the [es module support of the browser](https://caniuse.com/#feat=es6-module) which mean there is a lot of browsers version that might end up with the modern build without being fully supported by it.
And all browsers that get the legacy build except IE11 are not fully supported by the build either, since last 3 browsers are all compatible with es modules and will get the modern build...

And we have a simpler but similar mess with SC...

philippe-perrin-sonarsource: @wouter-admiraal-sonarsource Since it looks like the situation needs to be clearer, could you NOT use `finally` here for now ? @gregaubert and I are planning a task force to take the power back on this matter (at least to try :laughing: ).

helpers/scrolling.ts
wouter-admiraal-sonarsource: OK, will make `ScrollQueueItem` an interface 👍 .

PR #85, opened 2020-01-16
Update eslint plugins

gregaubert:


components/controls/buttons.css
```diff
@@ -185,6 +185,7 @@
 }
 
 /* #region .button-group */
+/* TODO drop usage of this class in SQ (already dropped from SC) */
```
philippe-perrin-sonarsource, line 188: @jeremy-davis-sonarsource @mathieu-suen-sonarsource Have we decided to move forward on this ?

helpers/__tests__/request-test.ts
```diff
@@ -95,7 +95,7 @@ describe('parseError', () => {
     ).resolves.toBe('Error1. Error2');
   });
 
-  it('should parse error and return default message', () => {
+  it('should parse error object and return default message', () => {
```
philippe-perrin-sonarsource: Not related to the PR: this test is incredibly difficult to understand IMO. Also, the loop only runs 2 times, so I'm not sure this really does test anything because the call after only is the third one. WDYT ?

helpers/__tests__/request-test.ts
```diff
@@ -233,7 +233,8 @@ describe('requestTryAndRepeatUntil', () => {
       stopRepeat
     );
 
-    expect(promiseResult).rejects.toBe(undefined);
+    // eslint-disable-next-line jest/valid-expect
+    expect(promiseResult).rejects.toBeUndefined();
```
philippe-perrin-sonarsource: Removing this line make the next test fail ... :upside_down_face:

components/charts/AdvancedTimeline.tsx
```diff
@@ -251,30 +251,33 @@ export default class AdvancedTimeline extends React.PureComponent<Props, State>
...
-            date.valueOf() - previousPoint.x.valueOf() <= nextPoint.x.valueOf() - date.valueOf())
-        ) {
-          idx--;
-        }
-        const selectedDate = firstSerie.data[idx].x;
-        const xPos = this.state.xScale(selectedDate);
-        this.setState({ selectedDate, selectedDateXPos: xPos, selectedDateIdx: idx });
-        if (updateTooltip) {
-          updateTooltip(selectedDate, xPos, idx);
+    this.setState(state => {
```
philippe-perrin-sonarsource, line 254: NP: you probably could continue to avoid calling the `setState` if conditions (L256) are not fulfilled. The `return null` at the end of the `setState` is weird IMO.

philippe-perrin-sonarsource (commented):
Thank you for taking the time to do implement those changes. I'm not giving you an approval because I think there is some issues with the `requests-test.ts` file. Happy to discuss it with you.
Also, you probably should ask for an SC's team member review.

helpers/__tests__/request-test.ts
gregaubert: Yeah I agree it's not the best way of testing promises. I could replace everything with some async/await.

components/charts/AdvancedTimeline.tsx
gregaubert: The `return null` is used because we don't want to update the state, I think we could also return an empty object and it would do the same.

I can't move the full condition of (L256) outside the `setState` since it's based on the state. But ok I could split the condition to avoid calling the `setState` if the `firstSerie` is `undefined`, but then it add one more layer of `if` around everything and just make things a bit more complex for probably not much performance gain (just avoiding a function call basically)...

jeremy-davis-sonarsource: Calling setState with null doesn't trigger a state change, so it's the basically the same as not calling it

helpers/__tests__/request-test.ts
gregaubert: I'm not sure which loop you are talking about ? It's in a different test right ?

components/controls/buttons.css
jeremy-davis-sonarsource: We'll tackle that when we work on them (or containing page). So, no plan for a plain refactoring ticket.

jeremy-davis-sonarsource (approved):
LGTM

helpers/__tests__/request-test.ts
philippe-perrin-sonarsource: The loop line 217. My understanding is that we want to leave room for more than 3 calls and check that only 3 calls happens. But I feel like we only leave room for 3 calls (2 in the loop, one after) and therefore we're not testing the edge case ('should stop after 3 calls').

gregaubert: Indeed, I fixed this by adding one more check at the end that it's not called anymore

philippe-perrin-sonarsource (approved):
LGTM

PR #86, opened 2020-01-16
SC-1258 Add copy to clipboard feature to ActionsDropdownItem

jonathan-weibel-sonarsource:
New features:
- Copy to clipboard from an ActionsDropdownItem;
- Popup positioning TopLeft.

They will be used in SonarCloud within the LineOptionsPopup, which open on source viewers, to copy line permalink / line content.

Deployment is needed for https://github.com/SonarSource/sonarcloud-core/pull/497

### DIFF
--- a/components/ui/popups.tsx
+++ b/components/ui/popups.tsx
@@ -22,12 +22,22 @@ import * as React from 'react';
 import ClickEventBoundary from '../controls/ClickEventBoundary';
 import './popups.css';
 
+/**
+ * Positioning rules:
+ * - Bottom = below the block, horizontally centered
+ * - BottomLeft = below the block, horizontally left-aligned
+ * - BottomRight = below the block, horizontally right-aligned
+ * - LeftTop = on the left-side of the block, vertically top-aligned
+ * - RightTop = on the right-side of the block, vertically top-aligned
+ * - TopLeft = above the block, horizontally left-aligned
+ */
 export enum PopupPlacement {
   Bottom = 'bottom',
   BottomLeft = 'bottom-left',
   BottomRight = 'bottom-right',
   LeftTop = 'left-top',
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "sonar-ui-common",
-  "version": "0.0.48",
+  "version": "0.0.49",
```
gregaubert: You should not update the version here yet, I'll show you how to release once the PR is ready.
On the other hand you should update the changelog already.

gregaubert (changes requested):
Just drop the change in the package.json and update the changelog and we will be good to go for me.

gregaubert (approved):
Looks good!

components/ui/popups.tsx
```diff
@@ -27,7 +27,8 @@ export enum PopupPlacement {
   BottomLeft = 'bottom-left',
   BottomRight = 'bottom-right',
   LeftTop = 'left-top',
-  RightTop = 'right-top'
+  RightTop = 'right-top',
+  TopLeft = 'top-left'
```
jeremy-davis-sonarsource, line 40: I find this a bit confusing, what's the difference between `left-top` and `top-left`?
I guess renaming is a pain, but a jsdoc comment could help?

jeremy-davis-sonarsource (approved):
LGTM

components/ui/popups.tsx
jonathan-weibel-sonarsource: Hi! I used the same logic than the other positioning, that is to say:

- **left-top** = the popup is positioned on the left side of the block, vertically aligned on top of the block;
- **top-left** = the popup is positioned above the block, horizontally aligned on the left of the block.

Not intuitive, I admit. I'll update the jsdoc.

PR #90, opened 2020-01-31
Fix duplicate lodash in builds

gregaubert:
This PR allows to remove lodash from being loaded in the initial chunks of sonar-web. And avoid having multiple copies of lodash duplicated in different chunks of the sonar-web build.

Related SC PR: https://github.com/SonarSource/sonarcloud-core/pull/539
Mainly the commit about [**Fix lodash duplication**](https://github.com/SonarSource/sonarcloud-core/pull/539/commits/67ddd94ec8a04d64237acd8f6192380ca6595385) is related to this change. The same should be applied to SQ build before updated to this version.

**Checklist**

* [ ] Functional validation
* [ ] Documentation update
* [x] Update changelog

### DIFF
--- a/helpers/cookies.ts
+++ b/helpers/cookies.ts
@@ -17,19 +17,24 @@
  * along with this program; if not, write to the Free Software Foundation,
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
-import { memoize } from 'lodash';
 
-const parseCookies = memoize(
-  (documentCookie: string): T.Dict<string> => {
-    const rawCookies = documentCookie.split('; ');
-    const cookies: T.Dict<string> = {};
-    rawCookies.forEach(candidate => {
-      const [key, value] = candidate.split('=');
-      cookies[key] = value;
-    });
-    return cookies;
+// Memoize only last parsed element
+const lastParsed: { cookie?: string; result?: T.Dict<string> } = {};
+
+function parseCookies(documentCookie: string): T.Dict<string> {
+  if (lastParsed.cookie === documentCookie && lastParsed.result) {
+    return lastParsed.result;
   }
-);
+  const rawCookies = documentCookie.split('; ');
+  const cookies: T.Dict<string> = {};
+  rawCookies.forEach(candidate => {
+    const [key, value] = candidate.split('=');
+    cookies[key] = value;
+  });
+  lastParsed.cookie = documentCookie;
+  lastParsed.result = cookies;
+  return cookies;
+}
 
 export function getCookie(name: string): string | undefined {
   return parseCookies(document.cookie)[name];
--- a/helpers/request.ts
+++ b/helpers/request.ts
@@ -17,7 +17,6 @@
  * along with this program; if not, write to the Free Software Foundation,
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
-import { isNil, omitBy } from 'lodash';
 import { stringify } from 'querystring';
 import { getCookie } from './cookies';
 import { translate } from './l10n';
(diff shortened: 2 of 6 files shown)


jonathan-weibel-sonarsource (approved):
LGTM! Actually, we could think about entirely replace lodash by our own functions.

helpers/request.ts
```diff
@@ -50,8 +49,14 @@ export function getCSRFToken(): T.Dict<string> {
 
 export type RequestData = T.Dict<any>;
 
-export function omitNil(obj: RequestData): RequestData {
-  return omitBy(obj, isNil);
+export function omitNil<T>(obj: T.Dict<T>): T.Dict<T> {
```
siegfried-ehret-sonarsource: The issue you have is not resolvable using imports from lodash/fp/... ?
If this piece of code is copy-pasted or adapted from lodash, how will we manage fixes and updates ? (well here it's easy to understand but we may do that for other functions).

tsconfig.json
```diff
@@ -6,7 +6,7 @@
     "noUnusedParameters": true,
     "strict": true,
     "strictFunctionTypes": false,
-    "module": "commonjs",
+    "module": "esnext",
```
siegfried-ehret-sonarsource, line 9: This works if we have this, right ?
https://github.com/SonarSource/sonarcloud-core/pull/539/files#diff-d1c6ddbef7eeec60f418b3572311d4daR72

siegfried-ehret-sonarsource (approved):

helpers/request.ts
gregaubert: Maybe it is, but I really wanted to get rid of it in this file.
This code is homemade and doesn't come from lodash, so no issue for maintaining it.

tsconfig.json
gregaubert: Exactly

gregaubert: In truth it can work without having this for all browsers that SonarCloud supports, but then we would end up with some duplicate copies of lodash...

gregaubert:
> LGTM! Actually, we could think about entirely replace lodash by our own functions.

Yeah we could, but I don't think we want to maintain such functions. I guess the important part is that only things we use from lodash is part of the build result.

philippe-perrin-sonarsource:
On our side, just making `esnext` the new module resolution in suc is enough to avoid having the whole `lodash` in our bundles:

module=commonjs:
image

module=esnext:
image

I'm wondering what it is on SC that makes you need to include suc in your babel build and rework those methods. Available to discuss that of course.

helpers/cookies.ts
```diff
@@ -17,19 +17,24 @@
...
-const parseCookies = memoize(
-  (documentCookie: string): T.Dict<string> => {
-    const rawCookies = documentCookie.split('; ');
-    const cookies: T.Dict<string> = {};
-    rawCookies.forEach(candidate => {
-      const [key, value] = candidate.split('=');
-      cookies[key] = value;
-    });
-    return cookies;
+// Memoize only last parsed element
```
jeremy-davis-sonarsource, line 21: why only the last?

helpers/request.ts
```diff
@@ -50,8 +49,14 @@ export function getCSRFToken(): T.Dict<string> {
...
+  return Object.keys(obj).reduce((newObj, key) => {
+    if (obj[key] == null) {
+      return newObj;
+    }
+    newObj[key] = obj[key];
+    return newObj;
+  }, {} as RequestData);
```
jeremy-davis-sonarsource: total NP:
```
if (obj[key] != null) {
  newObj[key] = obj[key];
}
return newObj;
```

jeremy-davis-sonarsource (approved):
only did a code review, no validation (philippe seems to have done that)

helpers/cookies.ts
gregaubert: We don't need more than this. We don't use cookies much so it's not so often that they change, they will usually do when you logout/login. But when you logout/login then the page is reloaded and the memoization cache is lost.
So in most case it's enough to just remember the last one, and if the cookie change without a page reload then we will cache it and it probably won't change anymore, and for sure won't go back to it's previous value (due to the nature of what we have in cookies).

gregaubert:
With the current implem we have trouble running tests in sonar-web, so I'll need to rework this PR...

philippe-perrin-sonarsource (changes requested):
Sonar-web tests are failing

PR #92, opened 2020-02-13
Feature/pper/radio

philippe-perrin-sonarsource:
FYI @wouter-admiraal-sonarsource

Also, please note that sonar-web has duplicate styles for `Radio` in `icons.css`. This should be removed to properly validate the change.

See https://github.com/SonarSource/sonar-enterprise/pull/2521/commits/3994dd38814d0fad01b1d878ba4eb7465df70d6c

### DIFF
--- a/components/controls/__tests__/Radio-test.tsx
+++ b/components/controls/__tests__/Radio-test.tsx
@@ -22,14 +22,31 @@ import * as React from 'react';
 import { click } from '../../../helpers/testUtils';
 import Radio from '../Radio';
 
-it('should render and check', () => {
+it('should render properly', () => {
+  const wrapper = shallowRender();
+  expect(wrapper).toMatchSnapshot('not checked');
+
+  wrapper.setProps({ checked: true });
+  expect(wrapper).toMatchSnapshot('checked');
+});
+
+it('should invoke callback on click', () => {
   const onCheck = jest.fn();
   const value = 'value';
-  const wrapper = shallow(<Radio checked={false} onCheck={onCheck} value={value} />);
-  expect(wrapper).toMatchSnapshot();
+  const wrapper = shallowRender({ onCheck, value });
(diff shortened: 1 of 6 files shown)


components/controls/__tests__/Radio-test.tsx
```diff
@@ -32,4 +32,9 @@ it('should render and check', () => {
   expect(onCheck).toBeCalledWith(value);
   wrapper.setProps({ checked: true });
   expect(wrapper).toMatchSnapshot();
+
+  onCheck.mockReset();
+  wrapper.setProps({ disabled: true });
+  click(wrapper);
+  expect(onCheck).not.toBeCalled();
```
jeremy-davis-sonarsource: NP: Create a separate test?

jeremy-davis-sonarsource (approved):
LGTM! (no functional validation)

gregaubert (approved):
Looking good and validated on SC :)

Just one suggestion: can we make the disabled state more visible ? Maybe by graying out the center of the radio dot too ? and not only the text ?

PR ready to be updated with the new version: https://github.com/SonarSource/sonarcloud-core/pull/577

components/controls/__tests__/Radio-test.tsx
philippe-perrin-sonarsource: I've clean them up a bit.

philippe-perrin-sonarsource:
> Just one suggestion: can we make the disabled state more visible ? Maybe by graying out the center of the radio dot too ? and not only the text ?

@gregaubert How's that sound ?

image

gregaubert:
Looks better indeed! Thanks! But isn't the label supposed to be grayed out too ?
Finally, do you think that when it's displayed like this and no option is selected yet, the user could think that the disabled option is the selected one ?

PR #93, opened 2020-02-20
Introduce primary button style

wouter-admiraal-sonarsource:
In SonarQube, we're (finally!) introducing the concept of a "primary button", which should distinguish itself as a primary call to action, compared to other buttons.

**Checklist**

* [x] Functional validation
* [ ] ~Documentation update~
* [x] Update changelog

### DIFF
--- a/components/controls/buttons.css
+++ b/components/controls/buttons.css
@@ -52,6 +52,15 @@
   box-shadow: 0 0 0 3px rgba(35, 106, 151, 0.25);
 }
 
+.button-primary {
+  background: var(--darkBlue);
+  color: #fff;
+}
+
+.button-primary:hover {
+  background: var(--veryDarkBlue);
+}
+
 .button.disabled,
 .button:disabled,
 .button:disabled:hover,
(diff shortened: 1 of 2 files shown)


components/controls/buttons.css
```diff
@@ -52,6 +52,15 @@
   box-shadow: 0 0 0 3px rgba(35, 106, 151, 0.25);
 }
 
+.button-primary {
+  background: var(--darkBlue);
+  color: #fff;
+}
+
+.button-primary:hover {
```
gregaubert, line 60: Should we also add a prop in the button component to toggle this ? So we don't have to know classes and mess with them too much when using a button ?

wouter-admiraal-sonarsource: I was thinking about that, yeah. Same for the `button-red` class, which is used all over the place. But I wasn't sure how to implement it, and this being kinda urgent, I went for the quick win 😄 .

gregaubert: Yeah fine with the quick win. Anyway I think the goal at some point will be to rewrite this whole buttons module to use emotion.

gregaubert (approved):

PR #94, opened 2020-02-20
Add Storybook to browse icons

siegfried-ehret-sonarsource:
I needed to browse the icons to look for something (and could not find it).
This is super quick and dirty but it is a start, hopefully we'll manage to improve that together 💝 .

I don't think it is worth mentioning in the changelog.

Screenshot 2020-02-26 at 15 30 42

### DIFF
--- a/stories/1-Icon.stories.tsx
+++ b/stories/1-Icon.stories.tsx
@@ -1,0 +1,331 @@
+/*
+ * Sonar UI Common
+ * Copyright (C) 2019-2019 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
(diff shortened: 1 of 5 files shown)


wouter-admiraal-sonarsource:
Wow, nice work. Lots of stuff you did there! I'm kind of wondering if/how this relates to the -- struggling/agonizing/almost-dead -- [Product Design System](https://github.com/SonarSource/products-design-system) 🤔 .

siegfried-ehret-sonarsource:
The Product Design what ? :p

stories/1-Icon.stories.tsx
```diff
@@ -0,0 +1,331 @@
...
+);
+export const Tree = () => <TreeIcon />;
+export const Treemap = () => <TreemapIcon />;
+export const Visible = () => <VisibleIcon />;
+export const Vulnerability = () => <VulnerabilityIcon />;
+export const Warning = () => (
+  <ThemeProvider theme={mockedTheme}>
+    <WarningIcon />
+  </ThemeProvider>
+);
```
jeremy-davis-sonarsource, line 331: I have to admit I'm a bit put-off by this story system, in which you have to import and export every single component.
Isn't there an easier way, and wrap all of them in the theme provider?

jeremy-davis-sonarsource (commented):
Interesting! I'm not sure this is the solution we really want, but it could be useful to dev other components as well.
We could maybe compare with a couple other similar products

stories/1-Icon.stories.tsx
siegfried-ehret-sonarsource: That's something we can discuss at the web-guild goûter !

philippe-perrin-sonarsource: I had the same feeling, arf, adding every component manually is such a pain.

gregaubert: Waw that's cool! But indeed it seems to be a bit of a pain to have to add them all manually, do you think it would be possible to mix this with tests ?
Or generate this file ?

siegfried-ehret-sonarsource:
Closing this for now, I could not manage to get a nicer solution. I'll try again later.

PR #95, opened 2020-02-21
Remove alert's tooltip in favor of an aria-label

philippe-perrin-sonarsource:
During the rewrite of the alert component, I've restored (it wasn't working) the tooltip that describes the type of the alert. It has been raised that it is kind of useless (and ugly ?), so I've dropped it in favor of an aria label to describe the alert to screen reader.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Remove alert's tooltip in favor of an aria-label
+
 ## 0.0.54
 
 - SC-1814 Add Radio styles to RadioCard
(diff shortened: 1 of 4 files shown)


wouter-admiraal-sonarsource (approved):
LGTM! Much better. Screen reader works nicely on macOS.

jonathan-weibel-sonarsource (approved):
Thank you for the ping. This makes me think I have noticed a similar issue with Modal: we often set the title of the modal in the contentLabel prop which is injected in the DOM as aria-label. We should use a description of the modal type instead.

philippe-perrin-sonarsource:
> Thank you for the ping. This makes me think I have noticed a similar issue with Modal: we often set the title of the modal in the contentLabel prop which is injected in the DOM as aria-label. We should use a description of the modal type instead.

@jonathan-weibel-sonarsource I've no idea what you're referring to, but I'll be happy to discuss it with you. Thanks for the review.

PR #96, opened 2020-02-24
Fix/ms/outside click mounting issue

mathieu-suen-sonarsource:
I happen to see a stacktrace regarding the OutsideClickHandler component due to unmount component. I have not be able to reproduce the issue but in case this happen again I think this will fix the issue.

**Checklist**

* [x] Functional validation
* [ ] ~~Documentation update~~
* [x] Update changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fixing unmount issue with click outside handler component.
+
 ## 0.0.52
 
 - Introduce primary button styles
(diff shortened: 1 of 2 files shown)


gregaubert (commented):
Seems good, I didn't do any test though...

wouter-admiraal-sonarsource (approved):
LGTM

jeremy-davis-sonarsource (approved):
Haven't tested, but it can't hurt, and :+1: for the cleanup

PR #97, opened 2020-03-02
SC-1733 Fix mouse event issues on disabled buttons

gregaubert:
**Checklist**

* [x] Functional validation
* [ ] ~Documentation update~
* [x] Update changelog

### DIFF
--- a/components/controls/__tests__/buttons-test.tsx
+++ b/components/controls/__tests__/buttons-test.tsx
@@ -19,7 +19,45 @@
  */
 import { shallow } from 'enzyme';
 import * as React from 'react';
-import { ButtonIcon, ButtonIconProps } from '../buttons';
+import { click, mockEvent } from '../../../helpers/testUtils';
+import { Button, ButtonIcon, ButtonIconProps } from '../buttons';
+
+describe('Button', () => {
+  it('should render correctly', () => {
+    const onClick = jest.fn();
+    const preventDefault = jest.fn();
+    const stopPropagation = jest.fn();
+    const wrapper = shallowRender({ onClick });
+    expect(wrapper).toMatchSnapshot();
+    click(wrapper.find('button'), mockEvent({ preventDefault, stopPropagation }));
+    expect(onClick).toBeCalled();
+    expect(preventDefault).toBeCalled();
+    expect(stopPropagation).not.toBeCalled();
+  });
+
+  it('should not stop propagation, but prevent default of the click event', () => {
+    const preventDefault = jest.fn();
+    const stopPropagation = jest.fn();
+    const wrapper = shallowRender({ preventDefault: false, stopPropagation: true });
+    click(wrapper.find('button'), mockEvent({ preventDefault, stopPropagation }));
+    expect(preventDefault).not.toBeCalled();
+    expect(stopPropagation).toBeCalled();
+  });
+
+  it('should disable buttons with a class', () => {
+    const onClick = jest.fn();
+    const button = shallowRender({ disabled: true, onClick }).find('button');
+    expect(button.props().disabled).toBeUndefined();
+    expect(button.props().className).toContain('disabled');
+    expect(button.props()['aria-disabled']).toBe(true);
+    click(button);
+    expect(onClick).not.toBeCalled();
+  });
+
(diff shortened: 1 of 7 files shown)


components/controls/buttons.tsx
```diff
@@ -74,8 +76,7 @@ export class Button extends React.PureComponent<ButtonProps> {
       // eslint-disable-next-line react/button-has-type
       <button
         {...props}
-        className={classNames('button', className)}
-        disabled={this.props.disabled}
```
siegfried-ehret-sonarsource, line 78: As discussed directly, I think we may have an issue since the disabled state is style only.
Can you add https://www.w3.org/WAI/PF/aria/states_and_properties#aria-disabled to bring a meaningful state to the button ? Thanks !

siegfried-ehret-sonarsource (approved):

components/controls/__tests__/buttons-test.tsx
```diff
@@ -19,7 +19,45 @@
...
+  it('should disable buttons with a class', () => {
+    const onClick = jest.fn();
+    const button = shallowRender({ disabled: true, onClick }).find('button');
+    expect(button.props().disabled).toBeUndefined();
+    expect(button.props().className).toContain('disabled');
+    expect(button.props()['aria-disabled']).toBe(true);
+    click(button);
+    expect(onClick).not.toBeCalled();
+  });
+
```
jeremy-davis-sonarsource, line 56: I'm wondering if we could/should test events on a disabled button, to cover the case that triggered this change in the first place?

jeremy-davis-sonarsource (approved):
LGTM, haven't tested though

components/controls/__tests__/buttons-test.tsx
gregaubert: It would be nice, I just tried, but wasn't able to make the test fail in jest in fact. And it seems legit since the issue happens only in Chrome currently...
So I don't know how to write such test so that it make sense, do you have ideas ?

gregaubert: What I tried:
```ts
  it('should still get mouseover and mouseleave events when disabled', () => {
    const onMouseOver = jest.fn();
    const onMouseLeave = jest.fn();

    const wrapper = shallow(
      <Button
        disabled={true}
        // @ts-ignore
        onFocus={jest.fn()}
        onMouseOver={onMouseOver}
        onMouseLeave={onMouseLeave}
      />
    );
    wrapper.find('button').simulate('mouseover');
    expect(onMouseOver).toBeCalled();
    wrapper.find('button').simulate('mouseleave');
    expect(onMouseLeave).toBeCalled();
  });
```

jeremy-davis-sonarsource: Ooh, `simulate('event')` is just a shortcut for `.props('onEvent')()`.. so yeah... that's never failing.
You would need to actually simulate the event like (thanks mdn!):
```
function simulateClick() {
  var evt = new MouseEvent("click", {
    bubbles: true,
    cancelable: true,
    view: window
  });
  var cb = document.getElementById("checkbox"); //element to click on
  var canceled = !cb.dispatchEvent(evt);
  if(canceled) {
    // A handler called preventDefault
    alert("canceled");
  } else {
    // None of the handlers called preventDefault
    alert("not canceled");
  }
}
```

gregaubert: Thanks! I'll try!
But does it even make sense to do that ? I mean, the bug is on chrome only no ?

jeremy-davis-sonarsource: Actually it seems it's not a bug, and React 16.13 explicitly gets rid of events on disabled buttons: https://reactjs.org/blog/2020/02/26/react-v16.13.0.html#notable-bugfixes

gregaubert: Okay so I gave it a new try and still no luck.
I tried generating real events with `dispatchEvent` like in your example and adding event listener to the dom node, but it's the same it always trigger even for disabled buttons.

I'm giving up on this for now so that I can move forward with my PR on SC side. We can still do it later, ping me if you have new ideas and want us to try together!

PR #98, opened 2020-03-03
SONAR-12060 Fix modal header for long contents

siegfried-ehret-sonarsource:
This allows to go from:

Screenshot 2020-03-03 at 09 44 53

To:

Screenshot 2020-03-03 at 09 44 36

**Checklist**

* [x] Functional validation
* [ ] Documentation update
* [x] Update changelog

### DIFF
--- a/components/controls/Modal.css
+++ b/components/controls/Modal.css
@@ -99,6 +99,7 @@
   font-size: var(--bigFontSize);
   font-weight: bold;
   line-height: normal;
+  overflow-wrap: break-word;
 }
 
 .modal-body {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Fixing unmount issue with click outside handler component.
+- SONAR-12060 Fix titles in modal for long words
 
 ## 0.0.52
 


jeremy-davis-sonarsource (approved):
LGTM

siegfried-ehret-sonarsource:
@jeremy-davis-sonarsource you LGTM ! <3

gregaubert (commented):
Cool that works, but it's only for `ConfirmModal` and `ValidationModal`. :cry: 
WDYT if we export a `ModalHead` component from our `Modal.tsx` instead. And that `ModalHead` is a styled component like you did that contains both the `head` and `h2` and also implement the `.modal-head` styling.
So that from now on when we see a modal in the SC/SQ code we can gradually start to replace all the:
```ts
<header className="modal-head">
   <h2>{header}</h2>
</header>
```
with:
```ts
<ModalHead>{header}</ModalHead>
```

And obviously our `ConfirmModal` and `ValidationModal` would use the `ModalHead`

wouter-admiraal-sonarsource (dismissed):
@gregaubert 's suggestion, although I prefer it in terms of code style, wouldn't take care of the header description 🤷‍♂. So not sure it's enough (unless we never use the description anywhere).

On my side, I find it very odd that we introduce a new component for the header, but none for the body or footer. Either we go the whole way, or we don't. Furthermore, I don't see much value in this refactor, as adding 3 lines of CSS would have done the trick:

```css
.modal-head h2 {
  overflow-wrap: break-word;
}
```

We can def. turn these into styled components, why not? But then, at the very least, we should add the footer and body as well, for consistency. And this would imply allowing them to have children (no way props would suffice), which in turn would mean the header component would look odd with props. So it would have to follow Greg's example. Which in turn would mean it cannot "force" a styled component for the H2 (because of the header description). Which means people would add an H2 tag manually. Which in turn means you would add the above mentioned CSS anyway. Which finally means the refactor would only serve to abstract away the `modal-head` CSS class that was added manually 😅.

In the end, I'm not sure it's worth it 🤷‍♂ .

I'm putting "Request changes", not because I want this to be updated per-se, but because I'd like @jeremy-davis-sonarsource and @gregaubert to reconsider their approval in the light of this. If the others still approve, I won't fight it, and I'm fine with merging it.

components/controls/Modal.css
```diff
@@ -88,11 +88,6 @@
   margin-right: var(--sbw);
 }
 
-.modal-head {
-  padding: calc(4 * var(--gridSize));
-  padding-bottom: 0;
-}
-
```
wouter-admiraal-sonarsource: This is a breaking change. It should not be removed, **unless** all modals are updated before updating sonar-ui-common. So, if this makes it into SUC, we (SC and SQ) would better make sure all our modals are ready for it!

gregaubert:
@wouter-admiraal-sonarsource Yeah I agree... If we do that with only the modal-head it becomes inconsistent with the rest... And I think it's worth to do it all the way, because we won't have to deal with knowing the dom for modals at some point and that will be good.

What's for sure if we want ot go this way is:
* we need to align on the api we want for those component
* we need to keep the old css classes to not break the code that uses it as long as we don't move everything to this new model, which I don't expect we will do in one go

siegfried-ehret-sonarsource:
I looove @wouter-admiraal-sonarsource 's view on this and I am sorry about the complexity of the previous thing. I followed your advice to keep things simple.

components/controls/Modal.css
```diff
@@ -101,6 +101,10 @@
   line-height: normal;
 }
 
+.modal-head h2 {
+  overflow-wrap: break-word;
+}
+
```
gregaubert: At least this is quick and simple fix for the current issue at hand!
I think you should even merge it to the previous class so that is apply to both `.modal-head h1, .modal-head h2`...

gregaubert (approved):

jeremy-davis-sonarsource (commented):
Ooh, yeah this has changed quite a bit.

I don't know if changing the `modal-head` to a styled-component is what we want.
Do we start writing styled-components for layout/styling that we then use in SQ/SC?
I'm personally not against it, but it should be a decision we make collectively.

Even if we do so, as was pointed out, the only Modals benefiting the change are the two in SUC, and not the 50+ ones in SQ/SC. Wouter's CSS version does fix the issue for all modals.

I vote for the change, but to also use the CSS fix.

PR #99, opened 2020-03-06
SC-1814 Add Radio styles to RadioCard

siegfried-ehret-sonarsource:
We broke some styles in [this](https://github.com/SonarSource/sonarcloud-core/commit/d294556c700a292ac14a971845f925457259b2f6), let's fix that !

The RadioCard was missing some styles from the Radio.

**Checklist**

* [ ] Functional validation
* [ ] Documentation update
* [ ] Update changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- SC-1814 Add Radio styles to RadioCard
+
 ## 0.0.53
 
 - Fixing unmount issue with click outside handler component.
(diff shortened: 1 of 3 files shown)


gregaubert (approved):
Looks good! I'm just wondering if we could use the `Radio` component directly inside the `RadioCard` ?

siegfried-ehret-sonarsource:
@gregaubert I looked into this way but did not move forward because the Radio introduces a `a` tag which is not what we want and it also needs a `onCheck` prop and I find it weird to have an element with a click handler inside an element with a click handler (the 2 handlers will be the same, but it does not sound good, right?).

gregaubert:
Ah yeah ok I see, then fine with current solution! :+1:

PR #101, opened 2020-03-18
SONAR-13155 Upgrade dependencies

jeremy-davis-sonarsource:
Goes with [SONAR#2615](https://github.com/SonarSource/sonar-enterprise/pull/2615)

/!\ SC will have to align some dependency versions for this to work!

**Checklist**

* [ ] Functional validation
* [x] Update changelog


components/charts/__tests__/__snapshots__/AdvancedTimeline-test.tsx.snap
```diff
@@ -39,8 +39,8 @@ exports[`should render correctly 1`] = `
           className="line-chart-grid"
           x1={0}
           x2={30}
-          y1={34.666666666666664}
-          y2={34.666666666666664}
+          y1={34.66666666666667}
```
siegfried-ehret-sonarsource, line 42: Interesting... 🤔

siegfried-ehret-sonarsource (approved):
Looks good to me, even if the +0.0000000004 on every value is strange.

components/charts/__tests__/__snapshots__/AdvancedTimeline-test.tsx.snap
jeremy-davis-sonarsource: My guess is that it's https://github.com/d3/d3-interpolate/issues/73 ?

PR #107, opened 2020-03-30
Fix for update-center's component

philippe-perrin-sonarsource:
I thought that using our getJSON utility was a good idea, but it turns out that it is located in a module that import 'l10n', which do some magic on the `window` object, magic that fails in the context of Gatsby ... So back to `window.fetch`.

### DIFF
--- a/components/ui/update-center/MetaData.tsx
+++ b/components/ui/update-center/MetaData.tsx
@@ -19,7 +19,6 @@
  */
 
 import * as React from 'react';
-import { getJSON } from '../../../helpers/request';
 import './MetaData.css';
 import MetaDataVersions from './MetaDataVersions';
 import { MetaDataInformation } from './update-center-metadata';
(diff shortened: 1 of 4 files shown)


components/ui/update-center/MetaData.tsx
```diff
@@ -55,14 +54,17 @@ export default class MetaData extends React.Component<Props, State> {
...
+          if (response.status >= 200 && response.status < 300) {
+            return response.json();
+          } else {
+            return Promise.reject(response);
           }
-        },
-        () => this.setState({ data: undefined })
-      );
+        })
+        .then(data => this.setState({ data }))
```
mathieu-suen-sonarsource: Maybe add `this.mounted` check ?

components/ui/update-center/MetaData.tsx
```diff
@@ -55,14 +54,17 @@ export default class MetaData extends React.Component<Props, State> {
...
+        .catch(() => this.setState({ data: undefined }));
```
mathieu-suen-sonarsource: same here

mathieu-suen-sonarsource (approved):
LGTM once you verify check is needed or not.

philippe-perrin-sonarsource:
> LGTM once you verify check is needed or not.

What ?

components/ui/update-center/MetaData.tsx
philippe-perrin-sonarsource: Good catch

components/ui/update-center/MetaData.tsx
philippe-perrin-sonarsource: Good catch

PR #108, opened 2020-04-14
SONAR-12256 Improve percentage formatting

wouter-admiraal-sonarsource:
Currently, we go through a lot of trouble to handle percentages rounding (see [this PR](https://github.com/SonarSource/sonar-enterprise/pull/2109)). This is an attempt to both fix this for PRs (where the issue still exists), and to simplify this whole approach.

The reasoning is, we round to a single decimal for "aesthetics", even though we always show 2 in failing conditions. Suggestion: let's always show 2, _unless_ there's a trailing zero, and trim it off. If there's a trailing double-zero, we only show 1 zero. So it should behave as expected, except it won't round by default anymore (always 2 decimals).

In this example, new coverage is at 91.69%, and overall is at 84.70%:

roundme

See https://github.com/SonarSource/sonar-enterprise/pull/2698 for the fix on sonar-enterprise.

**Checklist**

* [x] Functional validation
* [ ] ~Documentation update~
* [x] Update changelog

### DIFF
--- a/helpers/measures.ts
+++ b/helpers/measures.ts
@@ -173,14 +173,27 @@ function floatFormatter(value: string | number): string {
   return numberFormatter(value, 1, 5);
 }
 
-function percentFormatter(value: string | number, options: { decimals?: number } = {}): string {
+function percentFormatter(
+  value: string | number,
+  { decimals, omitExtraDecimalZeros }: { decimals?: number; omitExtraDecimalZeros?: boolean } = {}
+): string {
   if (typeof value === 'string') {
     value = parseFloat(value);
   }
-  if (options.decimals) {
-    return `${numberFormatter(value, options.decimals)}%`;
+  if (value === 100) {
+    return '100%';
+  } else if (omitExtraDecimalZeros && decimals) {
+    // If omitExtraDecimalZeros is true, all trailing decimal 0s will be removed,
+    // except for the first decimal.
+    // E.g. for decimals=3:
+    // - omitExtraDecimalZeros: false, value: 45.450 => 45.450
+    // - omitExtraDecimalZeros: true, value: 45.450 => 45.45
+    // - omitExtraDecimalZeros: false, value: 85 => 85.000
+    // - omitExtraDecimalZeros: true, value: 85 => 85.0
+    return `${numberFormatter(value, 1, decimals)}%`;
+  } else {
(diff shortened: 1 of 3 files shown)


helpers/measures.ts
```diff
@@ -173,14 +173,20 @@ function floatFormatter(value: string | number): string {
   return numberFormatter(value, 1, 5);
 }
 
-function percentFormatter(value: string | number, options: { decimals?: number } = {}): string {
+function percentFormatter(
+  value: string | number,
+  { decimals, singleTrailingZero }: { decimals?: number; singleTrailingZero?: boolean } = {}
```
jeremy-davis-sonarsource: NP: How about `omitExtraDecimalZeros`?

jeremy-davis-sonarsource (approved):
Nice work! This has been a thorn for a while

gregaubert (approved):
Looks good, and if I'm not mistaken it won't change the default behavior, so if we want to do the same we need to start setting the `singleTrailingZero` ?

helpers/measures.ts
wouter-admiraal-sonarsource: Wouldn't that be misleading? What is "extra"? For example, in 85.000, are the "extra" 0s all 3, or just the last 2? In our case, we want 85.000 to show as 85.0, not 85.

wouter-admiraal-sonarsource:
> @gregaubert Looks good, and if I'm not mistaken it won't change the default behavior, so if we want to do the same we need to start setting the `singleTrailingZero` ?

Indeed, it won't change the defaults. All other calls should not be affected.

helpers/measures.ts
wouter-admiraal-sonarsource: Ah, but then again, if we have 84.5900, and set `decimals` to 3 and `singleTrailingZero` to true, it will show 84.59, not 84.590, which might be expected from this flag 🤔 . But I'm not sure we'll ever run into such a case 😅 .

WDYT?

gregaubert: Whatever the name is, maybe a comment explaining exactly what will happen with this flag would make sense.

jeremy-davis-sonarsource: Exactly the point I was about to answer, and the case is much more common than the other. 90% of the time, statistically :smile:

PR #109, opened 2020-04-16
SONAR-13310 Add intl components

jeremy-davis-sonarsource:
Extracted the `intl` folder and added unit tests :metal: 
SQ/SC will be able to get rid of the duplicates in sonar-web

**Checklist**

* [x] Functional validation
* [ ] ~Documentation update~
* [x] Update changelog

### DIFF
--- a/components/intl/DateFormatter.tsx
+++ b/components/intl/DateFormatter.tsx
@@ -1,0 +1,40 @@
+/*
+ * Sonar UI Common
+ * Copyright (C) 2019-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import * as React from 'react';
+import { DateSource, FormattedDate } from 'react-intl';
+import { parseDate } from '../../helpers/dates';
+
+interface Props {
+  children?: (formattedDate: string) => React.ReactNode;
+  date: DateSource;
+  long?: boolean;
+}
+
+export const formatterOption = { year: 'numeric', month: 'short', day: '2-digit' };
+
+export const longFormatterOption = { year: 'numeric', month: 'long', day: 'numeric' };
+
+export default function DateFormatter({ children, date, long }: Props) {
+  return (
+    <FormattedDate value={parseDate(date)} {...(long ? longFormatterOption : formatterOption)}>
+      {children}
+    </FormattedDate>
+  );
+}
--- a/components/intl/DateFromNow.tsx
+++ b/components/intl/DateFromNow.tsx
@@ -1,0 +1,31 @@
+/*
+ * Sonar UI Common
+ * Copyright (C) 2019-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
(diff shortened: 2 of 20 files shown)


components/intl/DateFormatter.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+export interface DateFormatterProps {
+  children?: (formattedDate: string) => React.ReactNode;
+  date: DateSource;
+  long?: boolean;
+}
+
+export const formatterOption = { year: 'numeric', month: 'short', day: '2-digit' };
+
+export const longFormatterOption = { year: 'numeric', month: 'long', day: 'numeric' };
```
wouter-admiraal-sonarsource, line 32: How about SHOUTCASING (uppercasing) these? Or does that imply too much change in `sonar-web`?

components/intl/DateFromNow.tsx
```diff
@@ -0,0 +1,31 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import * as React from 'react';
+import { DateSource, FormattedRelative } from 'react-intl';
+import { parseDate } from '../../helpers/dates';
+
+interface Props {
+  children?: (formattedDate: string) => React.ReactNode;
```
wouter-admiraal-sonarsource, line 25: Really weird this is optional. Does it ever happen?

wouter-admiraal-sonarsource (approved):
LGTM! I take it these are just copy-pasted from sonar-web? (didn't check if anything seemed diff, and thus risks breaking).

components/intl/DateFromNow.tsx
jeremy-davis-sonarsource: about 50% of the time. It's when we don't want to wrap it in a node.

components/intl/DateFormatter.tsx
jeremy-davis-sonarsource: I think we shouldn't export these, and have a formatDate function that takes a `long` prop.. but I'm hesitant to do it

wouter-admiraal-sonarsource: In what cases are these used? I know I used it for the leak period components. Don't remember the details, though.

jeremy-davis-sonarsource: mostly when including a formatted date in a `translateWithParams` I think.
But I think we should use a `FormattedMessage` with the appropriate component rather than using these options manually

gregaubert: I looked quickly and the first one I found is in `CreationDateFacet`, and it wouldn't be easy (possible?) to switch for a component there...
The formatDate function seems like a good idea but I wonder how you'd get it to work outside react since it must use the one that is inside the intl context ?

components/intl/__tests__/DateFormatter-test.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { shallow } from 'enzyme';
+import * as React from 'react';
+import DateFormatter, { DateFormatterProps } from '../DateFormatter';
+
+it('should render correctly', () => {
+  expect(shallowRender()).toMatchSnapshot('standard');
+  expect(shallowRender({ long: true })).toMatchSnapshot('long');
+});
+
+function shallowRender(overrides: Partial<DateFormatterProps> = {}) {
```
gregaubert, line 29: Small tip: you don't need to export the props of functional component for testing anymore, you can do `Partial<typeof DateFormatter>` instead

gregaubert (approved):
Looks good! Thanks!

components/intl/__tests__/DateFormatter-test.tsx
jeremy-davis-sonarsource: I might be missing something, but wouldn't `CreationDateFacet` be "simply" fixed with:
```
const tooltip = (
        // eslint-disable-next-line react/jsx-fragments
        <React.Fragment>
          {formatMeasure(stats[start], 'SHORT_INT')}
          <br />
          <DateFormatter date={startDate} long={true} />
          {!isSameDay(tooltipEndDate, startDate) && (
            <DateFormatter date={tooltipEndDate} long={true}>
              {formatted => ` - ${formatted}`}
            </DateFormatter>
          )}
        </React.Fragment>
      );
```

gregaubert: No idea I didn't dig into it ^^

PR #110, opened 2020-04-22
Add init function for language bundle and web context

jonathan-weibel-sonarsource:
[Edit 2020-04-24]
These changes imply the initialization of the sonar-ui-common library before doing anything else with it. There are 3 available functions that can be called separately or chained: 
```
import Initializer from 'sonar-ui-common/helpers/init';

Initializer
    .setLocale('en')      // language
    .setMessages({})      // translation strings
    .setUrlContext('');   // [required] web context
```
So, the web context can be setup before the l10n data, since you might need to use the request helper to get the language bundle (on SQ side). And then you can setup the locale and messages with the language bundle itself.

You can also see that the theme is not initialized by these functions. We could have done some hacking like we do for the tests to initialize the theme through this functions but that would break the React Context, because we would not use the context provider anymore and changes of the theme object would not be reflected anymore in the context consumer (we don't do that currently but we prefer avoid surprises in the future).

Not initializing the lib at all will throw an error as soon as something related to the web context is used and write console warning in dev mode when something from the language bundles is used. We added a default English language bundle with only one translation for when a request fail, so in dev mode you might get some warnings when first initializing the lib with only the web context to get your language bundle and this request fail, we think it's acceptable.

Remarks and questions for reviewers:
* We decided to update some of our dependencies too since we are gonna jump version
* The update of prettier changed nearly all the files, making the coverage break, we tried to add some quick win tests to improve it but it's not enough and we don't plan to spend more time adding coverage for these changes
* Most interesting commit to review and that should deserve your attention is **SC-1951 Add sonnar-ui-common initialization function**
* Do you think it would be better to have separate initialize functions for the web context and the language bundle ? [x] --> now implemented
 

**Checklist**

* [x] Functional validation
* [x] Documentation update
* [x] Update changelog


helpers/__tests__/init-test.ts
```diff
@@ -0,0 +1,71 @@
...
+const originalConsoleError = console.error;
+console.warn = jest.fn();
+
+beforeEach(() => {
+  (console.warn as jest.Mock).mockClear();
+});
+
+afterAll(() => {
+  initSonarUICommon({ l10nLocale: 'en', l10nMessages: {}, urlContext: '' });
+  console.warn = originalConsoleError;
```
siegfried-ehret-sonarsource: You put `console.error` in `console.warn`. Is it on purpose ?

jonathan-weibel-sonarsource: OMG no! I switched to warn and forgot one replace. Thank you!

helpers/__tests__/init-test.ts
```diff
@@ -0,0 +1,71 @@
...
+  (console.warn as jest.Mock).mockClear();
+});
+
+afterAll(() => {
+  initSonarUICommon({ l10nLocale: 'en', l10nMessages: {}, urlContext: '' });
+  console.warn = originalConsoleWarn;
+});
+
+it('should throw when trying to get a value without initializing first', () => {
+  // @ts-ignore
```
siegfried-ehret-sonarsource: Can you make explicit why we need this like ? Like describe [here](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-6.html#suppress-errors-in-ts-files-using--ts-ignore-comments).

helpers/__tests__/l10n-test.ts
```diff
@@ -17,32 +17,33 @@
...
  */
-// require a "parent object" for spying
-import { requestMessages, resetBundle, translate, translateWithParameters } from '../l10n';
-import { getJSON } from '../request';
-
-jest.mock('../request', () => ({
-  getJSON: jest.fn(),
-}));
-
+/* eslint-disable camelcase */
```
siegfried-ehret-sonarsource, line 20: I removed this line and got no error when I ran eslint.

helpers/l10n.ts
```diff
@@ -17,49 +17,27 @@
...
-
-interface BundleRequestResponse {
-  effectiveLocale: string;
-  messages: LanguageBundle;
-}
-
-declare let window: Window & { SonarMessages: LanguageBundle };
-if (!window.SonarMessages) {
-  window.SonarMessages = {};
+export interface L10nMessages {
```
siegfried-ehret-sonarsource: Why not a `T.Dict<string>` ?

helpers/l10n.ts
```diff
@@ -17,49 +17,27 @@
...
 }
 
-export const DEFAULT_LANGUAGE = 'en';
-
 export function translate(...keys: string[]): string {
   const messageKey = keys.join('.');
-  if (process.env.NODE_ENV === 'development' && !window.SonarMessages[messageKey]) {
+  const l10nMessages = getL10nMessages();
+  if (process.env.NODE_ENV === 'development' && !l10nMessages[messageKey]) {
     // eslint-disable-next-line
```
siegfried-ehret-sonarsource:
```suggestion
    // eslint-disable-next-line no-console
```

helpers/l10n.ts
```diff
@@ -17,49 +17,27 @@
  * along with this program; if not, write to the Free Software Foundation,
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
-import { toNotSoISOString } from './dates';
-import { getJSON } from './request';
-import { get, save } from './storage';
+import { getL10nLocale, getL10nMessages } from './init';
```
siegfried-ehret-sonarsource: Maybe it is just me but `L10n` brings noise to the variables and functions names. WDYT ?

config/jest/SetupSUC.ts
```diff
@@ -0,0 +1,22 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import initSonarUICommon from '../../helpers/init';
+
+initSonarUICommon({ l10nLocale: 'en', l10nMessages: {}, urlContext: '' });
```
siegfried-ehret-sonarsource: This is fine but I would like to know if you thought about an API like this:
```
initSonarUICommon({ urlContext: '' })
  .configureMessages({locale: 'en', messages: {}}); 
```
I don't think it would bring any value now, but depending on how we want to make sonar-ui-common configurable it would be nice to use something to kind of "namespace" the configuration object into the API. 
That would also help removing the `l10n` noise I mention in another comment.

components/charts/__tests__/ColorGradientLegend-test.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { scaleLinear } from 'd3-scale';
+import ColorGradientLegend from '../ColorGradientLegend';
+import { shallow } from 'enzyme';
+// eslint-disable-next-line jest/no-mocks-import
+import { mockedTheme } from '../../__mocks__/mockedTheme';
+
+const { colors } = mockedTheme;
+const COLORS = [colors.green, colors.lightGreen, colors.yellow, colors.orange, colors.red];
+
+it('Should render properly', () => {
```
siegfried-ehret-sonarsource:
```suggestion
it('should render properly', () => {
```

components/charts/__tests__/ZoomTimeLine-test.tsx
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import { shallow } from 'enzyme';
+import * as React from 'react';
+// eslint-disable-next-line jest/no-mocks-import
+import { mockedTheme } from '../../__mocks__/mockedTheme';
```
siegfried-ehret-sonarsource, line 24: Would it be possible to disallow the `jest/no-mocks-import` rules globally for test files ?

components/controls/__tests__/ReloadButton-test.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+import * as React from 'react';
+import { shallow } from 'enzyme';
+// eslint-disable-next-line jest/no-mocks-import
+import { mockedTheme } from '../../__mocks__/mockedTheme';
+import { ThemeProvider } from '../../theme';
+import { click } from '../../../helpers/testUtils';
+import ReloadButton from '../ReloadButton';
+
+it('Should render properly', () => {
```
siegfried-ehret-sonarsource:
```suggestion
it('should render properly', () => {
```

components/ui/__tests__/GenericAvatar-test.tsx
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import * as React from 'react';
+import { shallow } from 'enzyme';
+import GenericAvatar from '../GenericAvatar';
+
+it('Should render properly', () => {
```
siegfried-ehret-sonarsource:
```suggestion
it('should render properly', () => {
```

components/controls/__tests__/BackButton-test.tsx
```diff
@@ -0,0 +1,39 @@
...
+ */
+import * as React from 'react';
+import { shallow } from 'enzyme';
+import { click } from '../../../helpers/testUtils';
+import { ThemeProvider } from '../../theme';
+// eslint-disable-next-line jest/no-mocks-import
+import { mockedTheme } from '../../__mocks__/mockedTheme';
+import BackButton from '../BackButton';
+
+it('Should render properly', () => {
```
siegfried-ehret-sonarsource:
```suggestion
it('should render properly', () => {
```

components/controls/__tests__/HelpTooltip-test.tsx
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+import * as React from 'react';
+import { shallow } from 'enzyme';
+// eslint-disable-next-line jest/no-mocks-import
+import { mockedTheme } from '../../__mocks__/mockedTheme';
+import { ThemeProvider } from '../../theme';
+import HelpTooltip from '../HelpTooltip';
+
+it('Should render properly', () => {
```
siegfried-ehret-sonarsource:
```suggestion
it('should render properly', () => {
```

components/controls/__tests__/BackButton-test.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const onClick = jest.fn();
+  const wrapper = shallow<BackButton>(<BackButton onClick={onClick} />, {
+    wrappingComponent: ThemeProvider,
+    wrappingComponentProps: {
+      theme: mockedTheme,
+    },
+  });
+  expect(wrapper).toMatchSnapshot();
+  click(wrapper.find('a'));
```
siegfried-ehret-sonarsource: The click would be better in another test, this one is only checking that it `should render properly`.

components/controls/__tests__/ReloadButton-test.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const onClick = jest.fn();
```
siegfried-ehret-sonarsource, line 36: Same here about the click to move to another test if possible.

components/controls/__tests__/SelectListListContainer-test.tsx
```diff
@@ -17,23 +17,34 @@
...
       onUnselect={jest.fn(() => Promise.resolve())}
+      readOnly={false}
       renderElement={(foo: string) => foo}
       selectedElements={['foo']}
     />
   );
+  await waitAndUpdate(wrapper);
   expect(wrapper).toMatchSnapshot();
+
+  // const mockHandle = jest.fn();
```
siegfried-ehret-sonarsource: Comments can be removed, right ?

components/controls/__tests__/SelectListListContainer-test.tsx
```diff
@@ -17,23 +17,34 @@
...
+  // wrapper.instance().handleBulkChange = mockHandle;
+  const cb = wrapper.find(Checkbox);
```
siegfried-ehret-sonarsource: Same here about the click to move to another test if possible.

components/controls/__tests__/ValidationModal-test.tsx
```diff
@@ -44,4 +47,7 @@ it('should render correctly', () => {
     </ValidationModal>
   );
   expect(wrapper).toMatchSnapshot();
+  wrapper.instance().handleSubmit(data);
```
siegfried-ehret-sonarsource: Same here about the submit to move to another test if possible.

components/intl/__tests__/DateTooltipFormatter-test.tsx
```diff
@@ -19,12 +19,22 @@
...
 import { shallow } from 'enzyme';
 import * as React from 'react';
+import { IntlProvider } from 'react-intl';
+import { getL10nLocale } from '../../../helpers/init';
 import DateTooltipFormatter from '../DateTooltipFormatter';
 
 it('should render correctly', () => {
-  expect(shallowRender()).toMatchSnapshot('standard');
+  const wrapper = shallowRender();
+  expect(wrapper).toMatchSnapshot();
```
siegfried-ehret-sonarsource:
```suggestion
  expect(shallowRender()).toMatchSnapshot();
```

siegfried-ehret-sonarsource (changes requested):

siegfried-ehret-sonarsource:
Great work on updating the deps and running prettier, thanks a lot! I am ok to not care about coverage now.

I added a few suggestions and a few questions. It is in "request changes" because I will run SonarCloud with this sonar-ui-common tomorrow.

helpers/__tests__/init-test.ts
gregaubert: Good idea! We want to have everything initialized to undefined in this test, it's not possible if we respect the types

README.md
```diff
@@ -23,6 +23,21 @@ Ambients types are available in the `types.d.ts`, to benefit from them in your p
 }
 ```
 
+To use the library it should first be initialized with by calling the following:
```
jeremy-davis-sonarsource: Typo: 
> To use the library it should first ~~with~~ be initialized by calling the following:

helpers/__tests__/l10n-test.ts
gregaubert: This is definitely needed, there is a lot of translation keys that don't respect the camelcase rule in this file, we don't want to have them reported on SQ.

This rule raise warnings, not errors, so you won't see them when running `yarn lint`

config/jest/SetupSUC.ts
siegfried-ehret-sonarsource: Thinking again about this, we could also inject the theme in some API.

components/charts/__tests__/ZoomTimeLine-test.tsx
gregaubert: I think this rule apply only to test files, it's from jest plugin. I think the issue here is more that this theme file should maybe not be in a mock folder, because it doesn't mock anything that exists in SUC. So it's not a mock, it's just a default theme used for tests, but I don't know where we should put it...

siegfried-ehret-sonarsource: Ok don't worry about it ! Thanks.

config/jest/SetupSUC.ts
gregaubert: Yeah there is probably way to improve the API, we tried to keep it simple there. But maybe it's better to have separate initialization function. But I feel like the API you suggested might be overly complex for what we need, it would mean return a class instance that has init functions that would set a global variable ?

Regarding the theme:
> You can also see that the theme is not initialized by this function, we could have do some hacking like we do for the tests to initialize the theme through this function but that would break the React Context, because we would not use the context provider anymore and changes of the theme object would not be reflected anymore in the context consumer (we don't do that currently but we prefer avoid surprises for the future).

gregaubert: I'd be happy to hear about the API from @jeremy-davis-sonarsource and @philippe-perrin-sonarsource too before we change something

helpers/l10n.ts
gregaubert: True, I'll drop the L10n part!

config/jest/SetupSUC.ts
siegfried-ehret-sonarsource: Simple is good ! :)
Yeah it is more complex, it would bring a class with methods that return the instance etc...
I'll just repeat one thing from my previous message: "I don't think it would bring any value now", there is no need to change anything but we can think about it during a Web coffee break !

helpers/__tests__/urls-test.ts
```diff
@@ -47,7 +42,7 @@ describe('#getPathUrlAsString', () => {
   });
 
   it('should take baseUrl into account', () => {
-    (window as any).baseUrl = '/context';
+    initSonarUICommon({ locale: 'en', messages: {}, urlContext: '/context' });
     expect(
       getPathUrlAsString({ pathname: '/dashboard', query: { id: COMPLEX_COMPONENT_KEY } })
     ).toBe('/context/dashboard?id=' + COMPLEX_COMPONENT_KEY_ENCODED);
```
jeremy-davis-sonarsource, line 48: NP: Add a test to check that `getHostUrl` throws as expected?

jeremy-davis-sonarsource (approved):
Wow, great job! This is going to be a good excuse to fix our bootstrapping code too...
Just a few NPs

Did a quick clutch to make it work with SQ, no issues detected

config/jest/SetupSUC.ts
jeremy-davis-sonarsource: I like the dissociation of the urlContext from the locale & messages. Maybe we could expose a plain object with references to individual functions ?
The setters could return the object, for chaining:

```
import initializer from 's-u-c/helpers/init'

initializer.setUrlContext('/');
initializer
  .setMessages({...})
  .setLocale('en');

```

jonathan-weibel-sonarsource: I like this design too, I'll give a try!

jonathan-weibel-sonarsource:
@gregaubert @jeremy-davis-sonarsource @siegfried-ehret-sonarsource I updated the initialization, using a small class with static methods. Could you have a look? Worked with SonarCloud.

jonathan-weibel-sonarsource:
> @gregaubert @jeremy-davis-sonarsource @siegfried-ehret-sonarsource I updated the initialization, using a small class with static methods. Could you have a look? Worked with SonarCloud.

The sample use in SC is in the third commit of this PR: https://github.com/SonarSource/sonarcloud-core/pull/861

siegfried-ehret-sonarsource:
Hey @jonathan-weibel-sonarsource maybe you don't need a class, and if you want one I'd do something like [this](https://github.com/SonarSource/sonar-ui-common/commit/fe1f05a6232642eef90f13f64b70000a2fc56812) (I haven't touched the tests and I don't know if it works ^^)

siegfried-ehret-sonarsource:
And it is also very fine how it was with only a function and a parameter object with everything, this is really an idea we can discuss and implement later.

jeremy-davis-sonarsource:
> Hey @jonathan-weibel-sonarsource maybe you don't need a class, and if you want one I'd do something like [this](https://github.com/SonarSource/sonar-ui-common/commit/fe1f05a6232642eef90f13f64b70000a2fc56812) (I haven't touched the tests and I don't know if it works ^^)

Yeah I was thinking no class as well just a plain:
```
const init = {
   setUrlContext = (url) => {urlContext = url ; return init; }
}
```

gregaubert:
Yeah I don't think we need a class, I'll simplify a bit :) Thanks for the feedback, I think it's better indeed with separate initializer!

gregaubert:
I'm gonna do a RC release of this so we can have the full QA running on SC side, but won't merge yet

gregaubert:
@philippe-perrin-sonarsource @jeremy-davis-sonarsource @siegfried-ehret-sonarsource I plan to do the actual release and merge this PR tomorrow (morning). So please raise the hand if you think there are some more changes needed or if you need more time to have a last look! :eyes:

philippe-perrin-sonarsource:
@gregaubert I need more time (was off last friday and busy with review/meetings today). Go ahead if you're fine with the review you've got so far. Otherwise, that's my top priority for tomorrow morning.

gregaubert:
> @gregaubert I need more time (was off last friday and busy with review/meetings today). Go ahead if you're fine with the review you've got so far. Otherwise, that's my top priority for tomorrow morning.

Great :+1: , I'll wait for your review tomorrow morning before merging (or not) then!

philippe-perrin-sonarsource (approved):
LGTM. Didn't tested it though. I think all the relevant comments has been made.

Good job guys.

PR #112, opened 2020-04-28
Add setReactDomContainer init function

gregaubert:
While switching to gatsby we noticed that this initialization function was missing. It allows to have a different selector for the dom node used to initialize the react app.
The default to "#content" is kept and no errors or warning are thrown by this one so that it's not needed to initialize for all other projects than SC.

**Checklist**

* [x] Functional validation
* [x] Documentation update
* [x] Update changelog

### DIFF
--- a/helpers/__tests__/init-test.ts
+++ b/helpers/__tests__/init-test.ts
@@ -23,6 +23,7 @@ import Initializer, {
   DEFAULT_MESSAGES,
   getLocale,
   getMessages,
+  getReactDomContainerSelector,
   getUrlContext,
 } from '../init';
 
@@ -34,7 +35,7 @@ beforeEach(() => {
 });
 
 afterAll(() => {
-  Initializer.setLocale('en').setMessages({}).setUrlContext('');
+  Initializer.setLocale('en').setMessages({}).setUrlContext('').setReactDomContainer(undefined);
   console.warn = originalConsoleWarn;
 });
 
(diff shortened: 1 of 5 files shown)


helpers/__tests__/init-test.ts
```diff
@@ -34,7 +35,7 @@ beforeEach(() => {
 });
 
 afterAll(() => {
-  Initializer.setLocale('en').setMessages({}).setUrlContext('');
+  Initializer.setLocale('en').setMessages({}).setUrlContext('').setReactDomContainer(undefined);
```
siegfried-ehret-sonarsource, line 38: TypeScript isn't angry here because the function definition is `setReactDomContainer(nodeSelector: string)` ?

siegfried-ehret-sonarsource: Just checked out the branch. TypeScript is angry ^^.

helpers/__tests__/init-test.ts
```diff
@@ -57,15 +58,26 @@ it('should throw when trying to get a value without initializing first', () => {
...
 
   expect(getLocale()).toBe(locale);
   expect(getMessages()).toBe(messages);
   expect(getUrlContext()).toBe(urlContext);
+  expect(getReactDomContainerSelector()).toBe(reactDomContainerSelector);
+  expect(console.warn).not.toBeCalled();
+});
+
+it('should have a default react dom container selector without warning', () => {
+  Initializer.setReactDomContainer(undefined);
```
siegfried-ehret-sonarsource, line 79: Here too.

siegfried-ehret-sonarsource (approved):
One typing to update then all good!

helpers/__tests__/init-test.ts
gregaubert: Heee... That's really weird that the validation passes indeed...

gregaubert: And why is it not angry on my side ???

gregaubert: That's really weird... On my machine it's angry if I try to put in a number or object but not for undefined or null...

siegfried-ehret-sonarsource: The `ts-check` task runs successfully on my machine, but I have the TypeScript error in my editor. 🤯 
IntelliJ uses the TypeScript version from sonar-ui-common, 3.8.3.

gregaubert: Got the same version too :exploding_head:

jeremy-davis-sonarsource (approved):
LGTM!

gregaubert:
Thanks for the review! I'll do the release!

PR #113, opened 2020-05-06
Avoid having the mock theme in the packaged bundle

gregaubert:
The mock theme has been bundled until now inside the package on npm and even up in the SC (and probably SQ) build too.

This PR removes it from there and move/rename the theme to the test config instead of mocks because it's not mocking anything.

**Checklist**

* [ ] Functional validation
* [ ] Documentation update
* [ ] Update changelog - didn't add anything since it should not have impact outside...

### DIFF
--- a/components/theme.ts
+++ b/components/theme.ts
@@ -26,9 +26,6 @@ import {
   withTheme,
 } from 'emotion-theming';
 import * as React from 'react';
-// mockedTheme is in fact the default theme
-// eslint-disable-next-line jest/no-mocks-import
-import { mockedTheme } from './__mocks__/mockedTheme';
 
 export interface Theme {
   colors: T.Dict<string>;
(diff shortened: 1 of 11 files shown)


siegfried-ehret-sonarsource (approved):

components/theme.ts
```diff
@@ -45,14 +42,6 @@ export interface ThemedProps {
 
 const ThemeContext = EmotionThemeContext as React.Context<Theme>;
 
-// Hack : override the default value of the context used for theme by emotion
-// As we can't pass a default value when it is constructed
-// This allows outside tests to get the mocked theme value without specifiying a theme provider
-if (process.env.NODE_ENV === 'test') {
-  (ThemeContext as any)['_currentValue'] = mockedTheme;
-  (ThemeContext as any)['_currentValue2'] = mockedTheme;
-}
```
philippe-perrin-sonarsource, line 54: I guess this has been introduced to avoid passing the context in each tests located in SC or SQ ("outside tests"), not only for SUC tests.

Are your SC tests green when linked to this version of SUC ? Sadly I've trouble to test it on the SQ side cause we've not implemented yet the new init process you introduced lately, and it is polluting my tests.

gregaubert: True! Good catch!
We need to update our "SetupTestEnvironment.js" file with something like this:
```js
const theme = require('../../src/app/theme');
const { default: ThemeContext } = require('sonar-ui-common/components/theme');

ThemeContext['_currentValue'] = theme;
ThemeContext['_currentValue2'] = theme;
```

I'll update the changelog and add this to the readme.

philippe-perrin-sonarsource: Please link the SC PR in this one. We'll try to follow you folks asap.

gregaubert: I didn't do a PR on SC yet, since this is not released, but I'll ping you on it once done

philippe-perrin-sonarsource (approved):
LGTM

gregaubert:
@philippe-perrin-sonarsource see https://github.com/SonarSource/sonarcloud-core/pull/934 for the changes on SC side

PR #114, opened 2020-05-12
Improve disabled select option style

antonio-garcia-sonarsource:
This change improves the visual style of disabled options in select elements with a better distinction between these and regular options.

This is related to this PR in sonarcloud: https://github.com/SonarSource/sonarcloud-core/pull/927

**Checklist**

* [ ] Functional validation
* [ ] Documentation update
* [ ] Update changelog

### DIFF
--- a/components/controls/Select.css
+++ b/components/controls/Select.css
@@ -339,7 +339,8 @@
 
 .Select-option.is-disabled {
   cursor: default;
-  opacity: 0.5;
+  opacity: 0.4;
+  font-style: italic;
 }
 
 .Select-noresults {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- SC-737 Improve disabled select option style
+
 ## 1.0.2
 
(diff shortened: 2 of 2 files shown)


gregaubert (changes requested):
Changes looks good, but you still need to update the changelog and you probably want to change the commit message that doesn't reflect what's done there.
(I guess it's the same commit message than on SC side, but it doesn't need to, just reference the ticket at the beginning and a meaningful text for the changes.)

gregaubert (approved):
Looks good to me!

components/controls/Select.css
```diff
@@ -339,7 +339,8 @@
 
 .Select-option.is-disabled {
   cursor: default;
-  opacity: 0.5;
+  opacity: 0.4;
+  font-style: italic;
```
jeremy-davis-sonarsource, line 343: .5 opacity already results in a failure in contrast rating, so I don't think it's a good idea to make it even more faded.

gregaubert: Laura suggested this change, because currently the contrast is too high on those disabled items and they don't seem like being disabled.
In any case I think we won't ever meet a correct contrast rating on that kind of disabled stuff, so it seems ok to me to lower it a bit more to make it more obvious that the item is disabled.

PR #116, opened 2020-05-29
Durcissement II: Behind the scenes

wouter-admiraal-sonarsource:
See https://github.com/SonarSource/sonar-enterprise/pull/2821

### DIFF
--- a/components/controls/buttons.css
+++ b/components/controls/buttons.css
@@ -276,9 +276,13 @@
   background-color: currentColor;
 }
 
-.button-icon:hover svg,
-.button-icon:focus svg {
+.button-icon:not(.disabled):hover svg,
+.button-icon:not(.disabled):focus svg {
   color: #fff;
+}
+
+.button.button-icon.disabled {
+  background: transparent !important;
 }
 
 /* #endregion */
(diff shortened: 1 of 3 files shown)


jeremy-davis-sonarsource (approved):
LGTM!

siegfried-ehret-sonarsource (approved):

wouter-admiraal-sonarsource:
Sorry, took the liberty of adding a 2nd commit. Could you have a look plz 🙏 ?

siegfried-ehret-sonarsource (approved):

components/controls/buttons.css
```diff
@@ -281,6 +281,10 @@
   color: #fff;
 }
 
+.button.button-icon.disabled {
+  background: transparent !important;
+}
```
jeremy-davis-sonarsource, line 286: Looks good, but the icon becomes white on hover, which means invisible in many cases. I'd make the icon always grey, but maybe it's better to add the background back on hover?

wouter-admiraal-sonarsource: Oh, good catch. I think keeping it gray makes more sense. Why have a hover state if it's disabled in the first place? Hover state suggests interactivity.

PR #118, opened 2020-06-10
SONAR-13391 Fix Activity Chart

jeremy-davis-sonarsource:
We added a new setting for the New Code which makes it possible for the leak to start before the first analysis. This breaks the Activity chart when the fork point is too old.
What happens is the leak `rect` becomes so big it generates some weird visual artifacts.

**Checklist**

* [x] Functional validation
* [ ] Documentation update
* [x] Update changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- SONAR-13391 Fix Activity Chart
+
 ## 1.0.5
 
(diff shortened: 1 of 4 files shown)


wouter-admiraal-sonarsource (approved):
LGTM! Did the functional validation as well.

gregaubert (approved):
Looks good! Well done!

PR #119, opened 2020-06-15
Add HttpStatus enum

siegfried-ehret-sonarsource:
Hello!

Every few months I go through some issues and I know we have a lot of "magic numbers".
This PR will allow to update some status codes to clean some of them in the products (obviously, this PR brings a lot of "magic numbers" 😅).

If you are a lucky reviewer here, that is because I added the suggested people from GitHub. 

Thanks!

**Checklist**

* [ ] Functional validation
* [ ] Documentation update
* [ ] Update changelog

### DIFF
--- a/helpers/request.ts
+++ b/helpers/request.ts
@@ -163,11 +163,11 @@ function checkApplicationVersion(response: Response): boolean {
 export function checkStatus(response: Response, bypassRedirect?: boolean): Promise<Response> {
   return new Promise((resolve, reject) => {
     if (checkApplicationVersion(response)) {
-      if (response.status === 401 && !bypassRedirect) {
+      if (response.status === HttpStatus.unauthorized && !bypassRedirect) {
         import('./handleRequiredAuthentication')
           .then((i) => i.default())
           .then(reject, reject);
-      } else if (response.status >= 200 && response.status < 300) {
+      } else if (isSuccessStatus(response.status)) {
         resolve(response);
       } else {
         reject(response);
@@ -236,7 +236,7 @@ export function getCorsJSON(url: string, data?: RequestData): Promise<any> {
     .setData(data)
     .submit()
     .then((response) => {
-      if (response.status >= 200 && response.status < 300) {
+      if (isSuccessStatus(response.status)) {
         return parseJSON(response);
       } else {
         return Promise.reject(response);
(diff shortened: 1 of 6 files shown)


helpers/request.ts
```diff
@@ -310,3 +313,69 @@ export function requestTryAndRepeatUntil<T>(
     }
   );
 }
+
+// Adapted from https://nodejs.org/api/http.html#http_http_HTTP_STATUS
+export enum HTTP_STATUS {
```
gregaubert: Cool! Should we rename it to `HttpStatus` to match the way we name other enums ? And usually we use `camelCase` for items in there.

gregaubert: Also maybe we should drop things that we don't use yet ? Or at least things that we might never use, I'm thinking about those 100s, or weird stuff like 203,205,206,207,226,...,418,...

Dropping theme will make the enum smaller, and make it easier to find what we want during autocompletion.

helpers/request.ts
```diff
@@ -236,7 +239,7 @@ export function getCorsJSON(url: string, data?: RequestData): Promise<any> {
     .setData(data)
     .submit()
     .then((response) => {
-      if (response.status >= 200 && response.status < 300) {
+      if (response.status >= HTTP_STATUS.OK && response.status < HTTP_STATUS.MULTIPLE_CHOICES) {
```
gregaubert: I think it because harder to understand like this, maybe we should still rely on 200 and 300 magic numbers introduce a helper function `isSuccessStatus` that those this check to not have them in too many places ?

gregaubert:
The magic numbers reported by SonarJS here are false positive, there is an improvement planned for this on their side: https://github.com/SonarSource/SonarJS/issues/1966

gregaubert (approved):

helpers/request.ts
```diff
@@ -310,3 +310,25 @@ export function requestTryAndRepeatUntil<T>(
     }
   );
 }
+
+export function isSuccessStatus(status: number) {
+  return status >= HttpStatus.ok && status < HttpStatus.multipleChoices;
```
jeremy-davis-sonarsource: I would keep the magic numbers here, because a success status = 2xx, not really that the status is between `ok` and `multipleChoices` (wat it mean ?? )

jeremy-davis-sonarsource (approved):
LGTM

helpers/request.ts
gregaubert: Yeah indeed

helpers/request.ts
philippe-perrin-sonarsource: Well, I'd have argued that camelCase is not the best here, given what we documented [here](https://github.com/SonarSource/sonar-enterprise/blob/master/private/docs/web-ui/index.md#types-and-enums) & everything that is [here](https://www.typescriptlang.org/docs/handbook/enums.html).

gregaubert: Ah yeah you are right indeed! Seems like we don't respect well the pascal case inside the enum yet on SonarCloud side!

siegfried-ehret-sonarsource: Thanks @philippe-perrin-sonarsource , I'll update that and release again :p

philippe-perrin-sonarsource: @siegfried-ehret-sonarsource Well, it works as is. Might not worth the effort of a release.

PR #121, opened 2020-06-17
Allow to pass aria-label to icons

gregaubert:
**Checklist**

* [ ] Functional validation
* [ ] Documentation update
* [ ] Update changelog


siegfried-ehret-sonarsource (approved):

aletsiram-castro-sonarsource (approved):
Yeah!!

components/icons/Icon.tsx
```diff
@@ -21,12 +21,14 @@ import * as React from 'react';
 import { Theme, ThemeConsumer } from '../theme';
 
 export interface IconProps {
+  'aria-label'?: string;
```
wouter-admiraal-sonarsource: As we're now basically allowing _any_ prop, I wonder if we should restrict to `aria-label` 🤔 ? Sometimes, `aria-hidden` could be interesting as well, for instance. Should we do something like:

```
export interface IconProps extends React.AriaAttributes {
```

Or even:

```
export interface IconProps extends React.HTMLAttributes<{}> {
```

?

wouter-admiraal-sonarsource (approved):
LGTM 👍 !

components/icons/Icon.tsx
gregaubert: Ah yeah good idea, I'll extends the `React.AriaAttributes`!

PR #124, opened 2020-06-19
WWWSC-86 Add possibility to have custom aria-label to Rating component

gregaubert:
We need custom aria-label on ratings for the marketing pages.

### DIFF
--- a/components/ui/Rating.tsx
+++ b/components/ui/Rating.tsx
@@ -23,16 +23,26 @@ import { translate, translateWithParameters } from '../../helpers/l10n';
 import { formatMeasure } from '../../helpers/measures';
 import './Rating.css';
 
-interface Props {
+interface Props extends React.AriaAttributes {
   className?: string;
   muted?: boolean;
   small?: boolean;
   value: string | number | undefined;
 }
 
-export default function Rating({ className, muted = false, small = false, value }: Props) {
+export default function Rating({
+  className,
+  muted = false,
+  small = false,
+  value,
+  ...ariaAttrs
+}: Props) {
   if (value === undefined) {
-    return <span aria-label={translate('metric.no_rating')}>–</span>;
+    return (
+      <span aria-label={translate('metric.no_rating')} {...ariaAttrs}>
+        –
+      </span>
+    );
   }
   const formatted = formatMeasure(value, 'RATING');
   return (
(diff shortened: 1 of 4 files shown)


components/ui/Rating.tsx
```diff
@@ -23,24 +23,30 @@ import { translate, translateWithParameters } from '../../helpers/l10n';
 import { formatMeasure } from '../../helpers/measures';
 import './Rating.css';
 
-interface Props {
+interface Props extends Pick<React.AriaAttributes, 'aria-label'> {
```
siegfried-ehret-sonarsource: Why only 'aria-label' and not other like for the icon in the previous PR?
(I am a noob in aria thing, not sure if other properties could be useful).

components/ui/__tests__/__snapshots__/Rating-test.tsx.snap
```diff
@@ -25,3 +25,20 @@ exports[`renders undefined value 1`] = `
   –
 </span>
 `;
+
+exports[`renders with a custom aria-label 1`] = `
+<span
+  aria-label="custom"
+  className="rating rating-B"
```
siegfried-ehret-sonarsource, line 33: Are we ok to have classes mixing lowercase and uppercase characters ?

siegfried-ehret-sonarsource (approved):
2 quick questions then go go go !

components/ui/__tests__/__snapshots__/Rating-test.tsx.snap
gregaubert: Ah yeah we could lower case these, but I'm a bit afraid to touch it, it's been like that for ages and I've no idea if it could be used somewhere else...

components/ui/Rating.tsx
gregaubert: I don't know, I just thought we only need aria-label so let's only keep this one, but like you I'm a noob at this...
@wouter-admiraal-sonarsource  ? :smile: Do you think it make sens to allow any aria attribute ?

wouter-admiraal-sonarsource: Well, ofc, not all ARIA attrs will make sense 🤷‍♂️ . But I'm not sure it's worth it to whitelist them. It's the same with any HTML: you can add any ARIA attribute to any element, but some of them won't make any sense, and will simply be ignored.

I'm personally OK with allowing any ARIA attribute, and let the dev make a call on what makes sense in what condition (e.g., I could imagine `aria-hidden` and `aria-labelledby` also make sense, depending on context).

gregaubert: I'll allow all of them then

aletsiram-castro-sonarsource (approved):
Zo/

jeremy-davis-sonarsource (commented):
I've got nothing to add :slightly_smiling_face:

PR #125, opened 2020-06-19
SONAR-13553 Do not allow multiple clicks on 'Load more' link in ListFooter

wouter-admiraal-sonarsource:
**Checklist**

* [x] Functional validation
* [ ] ~Documentation update~
* [x] Update changelog

### DIFF
--- a/components/controls/ListFooter.tsx
+++ b/components/controls/ListFooter.tsx
@@ -22,8 +22,9 @@ import * as React from 'react';
 import { translate, translateWithParameters } from '../../helpers/l10n';
 import { formatMeasure } from '../../helpers/measures';
 import DeferredSpinner from '../ui/DeferredSpinner';
+import { Button } from './buttons';
 
-export interface Props {
+export interface ListFooterProps {
   count: number;
   className?: string;
   loading?: boolean;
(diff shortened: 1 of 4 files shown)


components/controls/ListFooter.tsx
```diff
@@ -34,60 +34,61 @@ export interface Props {
...
+        onClick={(event: React.SyntheticEvent<HTMLElement>) => {
+          event.preventDefault();
+          event.currentTarget.blur();
+          if (props.reload) {
+            props.reload();
+          }
+        }}>
+        {translate('reload')}
+      </a>
+    );
```
jeremy-davis-sonarsource: Why not use a `ButtonLink`, make it disabled when `loading`?
It would simplify the callbacks significantly, too.

jeremy-davis-sonarsource (commented):
Matt wanted to get rid of buttons that look like links. Have you discussed this with him?
This is one of the components that appears the most with that meh-pattern :tm: 
--> maybe swap out the ButtonLink for a Button in my inline comment

components/controls/ListFooter.tsx
```diff
@@ -23,7 +23,7 @@ import { translate, translateWithParameters } from '../../helpers/l10n';
 import { formatMeasure } from '../../helpers/measures';
 import DeferredSpinner from '../ui/DeferredSpinner';
 
-export interface Props {
+export interface ListFooterProps {
```
gregaubert, line 27: You don't need to export those props, you can just do `typeof ListFooter` in the test to reference the props.

components/controls/ListFooter.tsx
```diff
@@ -34,60 +34,61 @@ export interface Props {
...
+      </a>
+    );
   } else if (hasMore && props.loadMore) {
-    link = loadMoreLink;
+    link = (
+      <a
+        className="spacer-left"
+        data-test="show-more"
+        href="#"
+        onClick={(event: React.SyntheticEvent<HTMLElement>) => {
```
gregaubert: I think `React.MouseEvent<HTMLAnchorElement>` is more appropriate.

gregaubert (approved):
Looks good!

components/controls/ListFooter.tsx
wouter-admiraal-sonarsource: Not quite. It kinda works for the `Partial<...>` argument to the `shallowRender()` function, because it functions almost like using `any` (you can pass any prop, even if incorrect). But more importantly, it doesn't work in the type passed to `shallow<T>()`:

```tsx
function shallowRender(props: Partial<typeof ListFooter> = {}) { // <-- This (kinda) works
  return shallow<typeof ListFooter>( // <-- This doesn't
    <ListFooter count={3} loadMore={jest.fn()} total={5} {...props} />
  );
}
```

In SQ, we try to type the return of the `shallow()` function, as it allows us to use the correct type in `state()` and `props()` calls.

For example, compare this:


Screenshot 2020-06-25 at 11 03 41

And this:


Screenshot 2020-06-25 at 11 05 07

Sure, it's kind of irrelevant in this case, but it's the guidelines we try to stick with in SQ 🙂 . I'm using TS 3.8.3 here.

Of course, this is a shared project with SC. Perhaps we want to align on this between teams? I'm not against using `typeof`, but I think we should make sure we understand pros and cons of each approach.

gregaubert: Wow... You are right! It doesn't work at all! I even wonder how we came to think that it works and that we should do that by default... Well I guess it's really flagrant when you do `wrapper.props().value` and we probably don't do that often for function component...

So it made me think and search a bit about it and I found this nice [`Parameters<T>` utility type](https://www.typescriptlang.org/docs/handbook/utility-types.html#parameterst), that means we could do :
```ts
function shallowRender(props: Partial<Parameters<typeof ListFooter>[0]> = {}) {
  return shallow<Parameters<typeof ListFooter>[0]>(
    <ListFooter count={3} loadMore={jest.fn()} total={5} {...props} />
  );
}
```

This works well, but it seems a bit tricky and long to write...

wouter-admiraal-sonarsource: Hmm, interesting. I wonder if we could create a custom type for this. Something like:

```ts
type PropsOf<T extends (...args: any) => any> = Parameters<typeof T>[0];
```

Unfortunately, the above doesn't work.

gregaubert: Yeah I tried something similar but didn't manage to make it work either...
But indeed, it would be great if we can come up with a custom type like this to not have this tricky notation!

wouter-admiraal-sonarsource:
@jeremy-davis-sonarsource Indeed, checked with Matt, and he agrees with the button. I updated the commit.

@gregaubert FYI, this will change all "Load more" links to buttons. This component is used in a _lot_ of places. I asked Matt to double check with Laura, and she's OK with this change. That being said, SC frontends still have some say in this matter. Do you accept this change (I've re-requested your review for this)?

gregaubert:
Yeah I'm fine with it. Did you have to update some places in SQ because it broke the layout ?

gregaubert (approved):

wouter-admiraal-sonarsource:
> Yeah I'm fine with it. Did you have to update some places in SQ because it broke the layout ?

Not that I noticed. I checked almost all places. It shouldn't really have a big impact, because the button is using `display: inline-flex;` by default. So it should only slightly increase the height of the footer, which is at the bottom of the UI anyway.

jeremy-davis-sonarsource (approved):

components/controls/ListFooter.tsx
gregaubert: This seems to work:
```ts
type PropsOf<T extends (...args: any) => any> = Parameters<T>[0];

function shallowRender(props: Partial<PropsOf<typeof ListFooter>> = {}) {
  return shallow<PropsOf<typeof ListFooter>>(
    <ListFooter count={3} loadMore={jest.fn()} total={5} {...props} />
  );
}
```
It's a little bit better than having to write the full thing but you still have to write the typeof...

wouter-admiraal-sonarsource: `Partial<PropsOf<typeof ListFooter>>` vs `ListFooterProps` ... I personally don't mind exporting the props interface and using it instead. Shorter, less error-prone, IMHO.

gregaubert: Yeah I agree that it's lighter and easier... I'm just not a fan that we have to export things just for the test, and always having to wonder when reviewing if these exported props are used/imported somewhere else where it shouldn't...
But I agree that for now it seems like the "best" solution...

PR #127, opened 2020-07-07
SC-2173 Introduce new component for SC new navigation

gregaubert:
Some upgraded popup to be able to use them anywhere like tooltips. Maybe I could have done it directly in popup but I was a bit too afraid of breaking too many stuff...

**Checklist**

* [ ] Functional validation
* [ ] Documentation update
* [ ] Update changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 - Add right-bottom popup placement
 - Allow to pass a ref to a Popup
 - Add PortalPopup component: allow to create a Popup outside it's overflow:hidden parent
+- Add animated Arrow icon
+- Add Cog icon
 
 ## 1.0.11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Add PortalPopup component: allow to create a Popup outside it's overflow:hidden parent
 - Add animated Arrow icon
 - Add Cog icon
+- Add AutoEllipsis component: allow to automatically add the .text-ellipsis class based on the text size
 
 ## 1.0.11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 ## Unreleased
 
 - SONAR-13530 Prevent default when button is disabled
+- Add right-bottom popup placement
+- Allow to pass a ref to a Popup
+- Add PortalPopup component: allow to create a Popup outside it's overflow:hidden parent
 
 ## 1.0.11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Add animated Arrow icon
 - Add Cog icon
 - Add AutoEllipsis component: allow to automatically add the .text-ellipsis class based on the text size
+- Add theme helpers to easily get theme values in emotion components
 
 ## 1.0.11
 
--- a/components/controls/Tooltip.css
+++ b/components/controls/Tooltip.css
@@ -23,9 +23,6 @@
   display: block;
   height: auto;
   box-sizing: border-box;
-}
-
-.tooltip {
   font-size: var(--smallFontSize);
   font-weight: 300;
   line-height: 1.5;
--- a/components/icons/ArrowIcon.tsx
+++ b/components/icons/ArrowIcon.tsx
@@ -1,0 +1,50 @@
+/*
+ * Sonar UI Common
+ * Copyright (C) 2019-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import * as React from 'react';
+import Icon, { IconProps } from './Icon';
+
+interface Props extends IconProps {
+  animated?: boolean;
+  inverseDirection?: boolean;
+}
+
+export default function ArrowIcon({
+  animated = false,
+  fill = 'currentColor',
+  inverseDirection = false,
+  ...iconProps
+}: Props) {
+  const style: React.CSSProperties = {};
+  if (inverseDirection) {
+    style.transform = 'scaleX(-1)';
+  }
+
+  if (animated) {
+    style.transition = 'transform 0.2s';
+  }
+  return (
+    <Icon style={style} {...iconProps}>
+      <path
+        d="M13.99 6.867l.668.005H4.99l3.04-3.046a.79.79 0 00.23-.561.789.789 0 00-.23-.56l-.473-.474A.784.784 0 006.998 2a.784.784 0 00-.558.23L1.23 7.44A.783.783 0 001 8c0 .212.081.41.23.56l5.21 5.21c.149.148.347.23.558.23.212 0 .41-.082.559-.23l.472-.473a.782.782 0 000-1.106L4.956 9.128H14a.819.819 0 00.801-.81v-.67c0-.435-.376-.78-.812-.78z"
+        fill={fill}
+      />
+    </Icon>
+  );
+}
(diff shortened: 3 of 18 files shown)


gregaubert:
@wouter-admiraal-sonarsource Pinging you because I think I saw that you worked quite a lot on the tooltips stuff

components/ui/PortalPopup.tsx
```diff
@@ -0,0 +1,228 @@
...
+              {this.renderActual}
+            </ScreenPositionFixer>
+          </PopupPortal>
+        )}
+      </>
+    );
+  }
+}
+
+class PopupPortal extends React.Component {
```
antonio-garcia-sonarsource: Maybe we can find a better name for this? I found confusing having PortalPopup and PopupPortal classes here. 

Otherwise, It looks good to me!

antonio-garcia-sonarsource (approved):

components/ui/PortalPopup.tsx
gregaubert: Ah yeah indeed, I didn't notice. Maybe I could rename it to something like: "PortalWrapper" or "PortalPopupWrapper", would it be better ? Or do you have something in mind?

antonio-garcia-sonarsource: Yes, I would say `PortalWrapper` fits perfectly here

gregaubert:
@antonio-garcia-sonarsource I also moved some new icons from SC to here.

@wouter-admiraal-sonarsource Do you think I should also move this component here: https://github.com/SonarSource/sonarcloud-core/pull/1132/commits/5977b4bf5b977d35c8bf98aa238a19843a7c1432 ?

components/ui/PortalPopup.tsx
```diff
@@ -0,0 +1,228 @@
...
+  width: number;
+}
+
+type State = Partial<Measurements>;
+
+function isMeasured(state: State): state is Measurements {
+  return state.height !== undefined;
+}
+
+export default class PortalPopup extends React.Component<PopupProps, State> {
```
siegfried-ehret-sonarsource: Not sure if that is useful but you can remove `children` from the PopupProps and use `react.PropsWithChildren<Props>` here.

gregaubert: Yes, I could but I don't want the children to be optional like in `React.PropsWithChildren`, so I just thought it was more straightforward to define it myself...

components/controls/Tooltip.css
```diff
@@ -23,9 +23,6 @@
   display: block;
   height: auto;
   box-sizing: border-box;
-}
-
-.tooltip {
```
wouter-admiraal-sonarsource, line 28: Bold move 😎 .

components/icons/ArrowIcon.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+interface Props extends IconProps {
+  animated?: boolean;
+  revert?: boolean;
+}
+
+export default function ArrowIcon({
+  animated = false,
+  fill = 'currentColor',
+  revert = false,
```
wouter-admiraal-sonarsource: NP: I'm not sure `revert` is intuitive here (I had to try it out to understand the intent). Maybe `leftToRight` would be more comprehensible?

components/ui/PortalPopup.tsx
```diff
@@ -0,0 +1,228 @@
...
+          top: this.state.top + topFix,
+          width: this.state.width,
+          height: this.state.height,
+        }
+      : undefined;
+
+    return (
+      <Popup
+        arrowStyle={
+          isMeasured(this.state)
```
wouter-admiraal-sonarsource: NP: You're calling this `isMeasured()` twice. You could also just check `style`:

```tsx
        arrowStyle={
         style
            ? this.adjustArrowPosition(placement, { leftFix, topFix })
            : undefined
        }
```

components/ui/PortalPopup.tsx
```diff
@@ -0,0 +1,228 @@
...
+          top = toggleRect.top + toggleRect.height;
+          break;
+        case PopupPlacement.BottomRight:
+          left = toggleRect.left + toggleRect.width - width;
+          top = toggleRect.top + toggleRect.height;
+          break;
+        case PopupPlacement.LeftTop:
+          left = toggleRect.left - width;
+          top = toggleRect.top;
+          break;
```
wouter-admiraal-sonarsource: Not sure all positionings are correct. See this for `LeftTop`:

Screenshot 2020-07-09 at 10 46 35

Could be due to the surrounding elements, which might confuse `ScreenPositionFixer`, but it's worth to double check.

components/ui/PortalPopup.tsx
```diff
@@ -0,0 +1,228 @@
...
+  width: number;
+}
+
+type State = Partial<Measurements>;
+
+function isMeasured(state: State): state is Measurements {
+  return state.height !== undefined;
+}
+
+export default class PortalPopup extends React.Component<Props, State> {
```
wouter-admiraal-sonarsource: NP: I wonder if this component shouldn't be added to the `popups.tsx` file, instead of living on its own. It's basically a more advanced variant of the `Popup` component, not its own thing. This feels a bit like the Buttons, which are all in the same file, regardless of their type.

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+- Add right-bottom popup position
+- Allow to pass ref to popup
```
wouter-admiraal-sonarsource: NP: 

> ...to Popup

Camelcase? So we know we're talking about the component class.

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
...
+- Add PortalPopup component: allow to create popup outside it's overflow:hidden parent
+- Add animated Arrow icon
+- Add cog icon
```
wouter-admiraal-sonarsource: NP: either lower case "Arrow", or camelcase "cog" 🙂 .

wouter-admiraal-sonarsource (approved):
LGTM 👍 . Some small nitpicks. Also, maybe double check the positioning works as intended.

wouter-admiraal-sonarsource:
> @wouter-admiraal-sonarsource Do you think I should also move this component here: [SonarSource/sonarcloud-core@5977b4b](https://github.com/SonarSource/sonarcloud-core/commit/5977b4bf5b977d35c8bf98aa238a19843a7c1432) ?

That looks pretty cool. Yes please 🙂 . Would be nice if we can get those tests to work, though. Need help (rubber duck) with that?

components/icons/ArrowIcon.tsx
gregaubert: WDYT about `inverseDirection` ?

components/ui/PortalPopup.tsx
gregaubert: I just checked again and I think it's correct. Here too, the top of the popup is on the left and is aligned with the top of your text. You'd get the same result if you used a normal popup.

gregaubert:
@wouter-admiraal-sonarsource , @antonio-garcia-sonarsource I applied all feedbacks and moved the autoellipsis component here and was able to cover it properly with tests. I also added some theme helpers to be used with emotion components.

antonio-garcia-sonarsource (approved):
LGTM!

gregaubert:
Ok I'll merge and proceed with the release so that I unlock my PR on SC side... @wouter-admiraal-sonarsource feel free to still add comments if you do a last review on the two last commit and find anything!

components/theme.ts
```diff
@@ -49,5 +49,17 @@ export const ThemeProvider = EmotionThemeProvider as React.ProviderExoticCompone
 >;
 export const useTheme = emotionUseTheme as () => Theme;
 
+export function themeGet(type: keyof Theme, name: keyof Theme['sizes']) {
```
wouter-admiraal-sonarsource, line 52: NP: technically, this `name` param should probably have been typed as `string` rather than `keyof Theme['sizes']` 🤷‍♂️ .

wouter-admiraal-sonarsource: And perhaps it doesn't need exporting?

components/ui/__tests__/AutoEllipsis-test.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+it('should render', () => {
+  const wrapper = shallow(
+    <AutoEllipsis maxWidth={5} useParent={false}>
+      <span className="medium">my test text</span>
+    </AutoEllipsis>
+  );
+
+  expect(wrapper).toMatchSnapshot();
+  expect(wrapper.find('span').hasClass('medium')).toBe(true);
```
wouter-admiraal-sonarsource, line 32: NP: I guess this one is redundant, if you have the snapshot.

wouter-admiraal-sonarsource:
Just added some NPs. And because it's already merged, feel free to ignore them 😉 .

components/theme.ts
gregaubert: It's good to have it exported, I use it already in a few places to get zIndexes. It's kind of the generic helper do get theme data we don't use often...

PR #128, opened 2020-07-13
SC-2174 Take the ContextNavBar z-index from the theme

gregaubert:
...description

**Checklist**

* [ ] Functional validation
* [ ] Documentation update
* [ ] Update changelog

### DIFF
--- a/components/icons/RocketIcon.tsx
+++ b/components/icons/RocketIcon.tsx
@@ -1,0 +1,32 @@
+/*
+ * Sonar UI Common
+ * Copyright (C) 2019-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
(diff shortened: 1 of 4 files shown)


config/jest/testTheme.ts
```diff
@@ -153,19 +153,23 @@ export default {
...
 
-    // ui elements
+    // page elements
     pageMainZIndex: '50',
 
-    tooltipZIndex: '8000',
-
-    dropdownMenuZIndex: '7500',
+    // generic page fixed elements
+    contextbarZIndex: '420',
```
gregaubert, line 160: I added this missing one and re-ordered a bit the list to match the comment on top.

components/ui/ContextNavBar.css
```diff
@@ -20,7 +20,7 @@
 .navbar-context,
 .navbar-context .navbar-inner {
   background-color: #fff;
-  z-index: 420;
+  z-index: var(--contextbarZIndex);
```
gregaubert, line 23: I was hesitant to not remove the line with `z-index: 420;` and keep both, like this:
```css
  z-index: 420;
  z-index: var(--contextbarZIndex);
```
So that is would not be a breaking change for SQ, wdyt ?

components/icons/RocketIcon.tsx
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright (C) 2009-2020 SonarSource SA
+ * All rights reserved
+ * mailto:info AT sonarsource DOT com
+ */
```
wouter-admiraal-sonarsource: This will fail the license check, I gather 🙂 . If not... we should update license check.

components/ui/ContextNavBar.css
wouter-admiraal-sonarsource: It would probably be confusing... We'd at least need a comment in that case. But I don't think it's necessary. The Z-index remains the same value, after all.

wouter-admiraal-sonarsource (approved):
LGTM 👍!

components/icons/RocketIcon.tsx
gregaubert: Arf, I always do that when moving things from SC repo to SUC...

components/ui/ContextNavBar.css
gregaubert: Yeah indeed, so you'll have to think about adding this value to your theme file when you update to this version of SUC.

PR #131, opened 2020-07-21
Add round Avatar optional prop

antonio-garcia-sonarsource:
Sonarcloud is going to start using completely rounded avatars in a couple of places inside the New UI. In order to make this easy and clean I added an optional round prop to the GenericAvatar component.

**Checklist**

* [ ] Functional validation
* [ ] Documentation update
* [x] Update changelog

### DIFF
--- a/components/ui/GenericAvatar.tsx
+++ b/components/ui/GenericAvatar.tsx
@@ -24,10 +24,11 @@ import { getTextColor, stringToColor } from '../../helpers/colors';
 interface Props {
   className?: string;
   name: string;
+  round?: boolean;
   size: number;
 }
 
-export default function GenericAvatar({ className, name, size }: Props) {
+export default function GenericAvatar({ className, name, round, size }: Props) {
   const color = stringToColor(name);
 
   let text = '';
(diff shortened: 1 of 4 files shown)


components/ui/GenericAvatar.tsx
```diff
@@ -43,6 +44,7 @@ export default function GenericAvatar({ className, name, size }: Props) {
       className={classNames(className, 'rounded')}
       style={{
         backgroundColor: color,
+        borderRadius: round ? '50%' : undefined,
```
gregaubert, line 47: does the `undefined` value override the `.rounded` class radius ?

gregaubert: It would also be good to add a test for that new prop

gregaubert (changes requested):
Looks good!
Just add the missing test, and confirm that the `undefined` value doesn't break the rounded class

components/ui/GenericAvatar.tsx
antonio-garcia-sonarsource: No problem regarding the `.rounded` class border-radius as the element border-radius is not even there when using undefined. I just tested it and there's no problem there.

gregaubert: Nice :)

antonio-garcia-sonarsource: I just added the missing test 👍

gregaubert (approved):

PR #134, opened 2020-08-04
SC-2266 Update page refresh behavior

siegfried-ehret-sonarsource:
...description

**Checklist**

* [ ] Functional validation
* [ ] Documentation update
* [ ] Update changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Bump to eslint-config-sonarqube@0.6.1
+- SC-2266 Add a `setRequestOptions`  to the initialisation process
 
 ## 1.0.15
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+## 1.0.16
+
 - Bump to eslint-config-sonarqube@0.6.1
 - SC-2266 Add a `setRequestOptions`  to the initialisation process
 
--- a/helpers/init.ts
+++ b/helpers/init.ts
@@ -18,11 +18,13 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 import type { Messages } from './l10n';
+import { RequestOptions } from './request';
 
 let urlContext: string; // Is the base path (web context) in SQ
 let messages: Messages | undefined;
 let locale: string | undefined;
 let reactDomContainerSelector: string | undefined; // CSS selector of the DOM node where the React App is attached
+let requestOptions: RequestOptions = {};
 
 export const IS_SSR = typeof window === 'undefined';
 export const DEFAULT_LOCALE = 'en';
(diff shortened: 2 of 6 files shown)


helpers/request.ts
```diff
@@ -149,11 +150,27 @@ export function corsRequest(url: string, mode: RequestMode = 'cors'): Request {
...
 function checkApplicationVersion(response: Response): boolean {
   const version = response.headers.get('Sonar-Version');
   if (version) {
     if (currentApplicationVersion && currentApplicationVersion !== version) {
-      window.location.reload();
+      if (getRequestOptions().refreshBehavior === RefreshBehavior.NotifyWaitReload) {
+        notifyWaitReload();
+      } else {
+        reload();
+      }
```
gregaubert: I think that the whole part about the notification center is going a bit too far into putting pure SC stuff into SUC, and makes things more complex.
What about only adding the request options part in the init, and the option would just be a `versionReload` function that is called instead of doing `window.location.reload()` ? That request option would be completely optional, and the default if not provided would be to do like before.

I think it would simplify a bit and we would have less dependencies on SUC for the implem in SC, and it would still allow us to do whatever we want.

helpers/types.ts
```diff
@@ -20,3 +20,9 @@
 export function isDefined<T>(x: T | undefined | null): x is T {
   return x !== undefined && x !== null;
 }
+
+// Request types
+
+export type RequestOptions = {
+  onVersionChange?: VoidFunction;
+};
```
gregaubert: I'm not sure it's the right file to put a type because it's inside the helper folder, so I'd say it should only contain helper functions...

I guess this type would be fine inside `init.ts`...

gregaubert (commented):
Looks nice and simple now!

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Bump to eslint-config-sonarqube@0.6.1
+- SC-2266 Add a `setRequestOptions`  to the initialisation process, with an object of options (currently only `onVersionChange?: VoidFunction;` is available).
```
philippe-perrin-sonarsource: NP: I don' think you need to describe the shape of the options in the changelog :wink:

helpers/init.ts
```diff
@@ -19,10 +19,15 @@
  */
 import type { Messages } from './l10n';
 
+export type RequestOptions = {
```
philippe-perrin-sonarsource: NP: Why not just an `interface`?

philippe-perrin-sonarsource: NP: Also, it will be more consistent if this type was in `request.ts` where it is consumed. See `Messages` type location for example.

philippe-perrin-sonarsource (approved):
LGTM. It seems that everything you added is optional (to ease our SQ version bump). Do you confirm you wrote it this way?

siegfried-ehret-sonarsource:
Hey @philippe-perrin-sonarsource yes it is intended to keep it simple and not breaking for you 😘 .

PR #136, opened 2020-08-25
SC-2266 onVersionChange request option returns if request can proceed

gregaubert:
On SC we want to let users keep using the app even if the api version changed, that's why we need the onVersionChange request option to also handle this

**Checklist**

* [ ] Functional validation
* [ ] Documentation update
* [ ] Update changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+## 1.0.17
+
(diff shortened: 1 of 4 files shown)


gregaubert:
FYI @siegfried-ehret-sonarsource

philippe-perrin-sonarsource (approved):

PR #137, opened 2020-08-25
Display the full date when hovering a date from now

philippe-perrin-sonarsource:
This add a title with the full date to all usage of `DateFromNow`.

**Breaking changes:** this PR removes the component `DateTooltipFormatter`: it was used only twice and for nothing according to our discussion on [tooltip/title/date](https://sonarsource.slack.com/archives/CKQPPFBL3/p1598369103000800). See [this](https://github.com/SonarSource/sonarcloud-core/search?q=DateTooltipFormatter&unscoped_q=DateTooltipFormatter) list.

See this [PR](https://github.com/SonarSource/sonar-enterprise/pull/3081) on the SonarQube side.

**Checklist**

* [ ] Functional validation
* [ ] Documentation update
* [ ] Update changelog


components/intl/DateFromNow.tsx
```diff
@@ -20,12 +20,31 @@
...
 
-export default function DateFromNow({ children, date }: Props) {
-  return <FormattedRelative value={parseDate(date)}>{children}</FormattedRelative>;
+export default function DateFromNow(props: DateFromNowProps) {
+  const { children = (f: string) => f, date } = props;
+
+  if (!date) {
+    // eslint-disable-next-line react/jsx-no-useless-fragment
+    return <>{children(translate('never'))}</>;
+  }
```
jeremy-davis-sonarsource:
```
if (!date) {
    return children(translate('never'));
  }
```
?

components/intl/DateFromNowHourPrecision.tsx
```diff
@@ -21,36 +21,20 @@ import { differenceInHours } from 'date-fns';
...
 
-  /*
-   * Cast returns as ReactElement to bypass typescript issue with functional components
-   * returning anything else (like a string, in our case...)
-   * (https://github.com/DefinitelyTyped/DefinitelyTyped/issues/20544)
-   */
-  if (!date) {
-    return children(translate('never')) as React.ReactElement;
+  if (date && differenceInHours(Date.now(), date) < 1) {
+    children = () => originalChildren(translate('less_than_1_hour_ago'));
```
jeremy-davis-sonarsource: I think this is kind of complicated.
If this triggers, then `DateFromNow` gets a children callback that ignores its param, so basically it will go through all the date formatting code for nothing.
The only benefit is the recycling of the structure, but that's just `<span title={formattedDate}>`

Why not make the hour precision a prop of `DateFromNow`, and add this condition to it?

jeremy-davis-sonarsource (commented):
This is great! We can discuss the point I raised if you want

components/intl/DateFromNowHourPrecision.tsx
gregaubert: I agree it's quite complicated, I think even duplicating the span title thing from `DateFromNow` is fine in this case, just returning:
```ts
    <DateTimeFormatter date={parseDate(date)}>
      {(formattedDate) => (
        <span title={formattedDate}>
          {translate('less_than_1_hour_ago')}
        </span>
      )}
    </DateTimeFormatter>
```

philippe-perrin-sonarsource: @gregaubert I guess you're fine with dropping the `DateFromNowHourPrecision` component then?

components/intl/DateFromNow.tsx
philippe-perrin-sonarsource: Sadely no. That would make the component to possibly return something else than an `JSX.Element`, and causes issues where the component is consumed.

components/intl/DateFromNowHourPrecision.tsx
gregaubert: I'm not sure what you mean, I think I was talking about `DateFromNowHourPrecision` implem in my comment.

philippe-perrin-sonarsource: Also there's another benefit here: the message is passed to the children so it can be processed. Still I agree this is complicated. I'll try to simplify it.

philippe-perrin-sonarsource:
```ts
<DateFromNow date={user.lastConnectionDate} hourPrecision={true}>
    {date => "TOTO:" + date}
</DateFromNow>
```

wouldn't work anymore in case of _never_ or _< 1 hour_

philippe-perrin-sonarsource: > I'm not sure what you mean, I think I was talking about DateFromNowHourPrecision implem in my comment.

I was referring to @jeremy-davis-sonarsource comment:

> Why not make the hour precision a prop of DateFromNow

gregaubert: Ahh ok got it! Yeah I'm fine with it!

gregaubert (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- SONAR-13796 Display the full date when hovering a date from now
```
gregaubert, line 5: Please just also list in the changelog the components that where dropped (and eventually by what they should be replaced)

PR #139, opened 2020-09-14
Drop deprecated react features + update lodash

gregaubert:
**Checklist**

* [ ] Functional validation
* [ ] Documentation update
* [ ] Update changelog


components/ui/DeferredSpinner.tsx
```diff
@@ -79,6 +81,11 @@ export default class DeferredSpinner extends React.PureComponent<Props, State> {
         )
       );
     }
-    return this.props.children || null;
+    return (
+      this.props.children ||
+      (this.props.placeholder ? (
+        <i className={classNames('deferred-spinner-placeholder', this.props.className)} />
+      ) : null)
```
jeremy-davis-sonarsource, line 88: What's the use case for this?

gregaubert: Use case is to stop using this class alone and in different ways in the SC codebase. So once this version is released I'll replace all the class usage with the `placeholder={true}` prop

siegfried-ehret-sonarsource (approved):

jeremy-davis-sonarsource (approved):

PR #140, opened 2020-09-16
SONAR-13758 Add new code legend to Activity Chart

jeremy-davis-sonarsource:
New legend for new code in the activity graph. Added a boolean prop to activate it.

**Checklist**

* [x] Functional validation
* ~Documentation update~
* [x] Update changelog

### DIFF
--- a/components/charts/AdvancedTimeline.tsx
+++ b/components/charts/AdvancedTimeline.tsx
@@ -24,7 +24,7 @@ import { area, curveBasis, line as d3Line } from 'd3-shape';
 import { flatten, isEqual, sortBy, throttle, uniq } from 'lodash';
 import * as React from 'react';
 import { isDefined } from '../../helpers/types';
-import { ThemeConsumer } from '../theme';
+import { Theme, ThemeConsumer } from '../theme';
 import './AdvancedTimeline.css';
 import './LineChart.css';
 
@@ -32,6 +32,7 @@ export interface Props {
   basisCurve?: boolean;
   endDate?: Date;
   disableZoom?: boolean;
+  displayNewCodeLegend?: boolean;
   formatYTick?: (tick: number | string) => string;
   hideGrid?: boolean;
   hideXAxis?: boolean;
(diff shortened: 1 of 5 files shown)


components/charts/AdvancedTimeline.tsx
```diff
@@ -56,7 +57,12 @@ type XScale = ScaleTime<number, number>;
 // TODO it should be `ScaleLinear<number, number> | ScalePoint<number> | ScalePoint<string>`, but it's super hard to make it work :'(
 type YScale = any;
 
+const LEGEND_PADDING = 4;
+const LEGEND_BOTTOM_MARGIN = 4;
```
gregaubert: Why not use the theme grid size instead ?

components/charts/AdvancedTimeline.tsx
```diff
@@ -351,8 +357,47 @@ export default class AdvancedTimeline extends React.PureComponent<Props, State>
...
+          height={LEGEND_LINE_HEIGHT}
+          width={legendWidth}
+          x={legendBackgroundPosition}
+          y={yRange[yRange.length - 1] - LEGEND_LINE_HEIGHT - legendMargin}
+        />
+        <text
+          className="new-code-legend"
+          ref={(node) => {
+            this.setState({ leakLegendTextWidth: node?.getBoundingClientRect().width });
+          }}
```
gregaubert: I don't know if it's important or not but you might want to extract the inlined function to avoid this caveats with callback refs:
https://reactjs.org/docs/refs-and-the-dom.html#caveats-with-callback-refs

Of put an if around your `setState` if the node isn't set.

components/charts/AdvancedTimeline.tsx
```diff
@@ -365,20 +410,24 @@ export default class AdvancedTimeline extends React.PureComponent<Props, State>
...
         {(theme) => (
-          <rect
-            className="leak-chart-rect"
-            fill={theme.colors.leakPrimaryColor}
-            height={yRange[0] - yRange[yRange.length - 1]}
-            width={leakWidth}
-            x={leakStart}
-            y={yRange[yRange.length - 1]}
-          />
+          <>
```
gregaubert, line 435: Should it be a svg group: `<g>` ?

gregaubert (commented):
Code looks good, just a few nitpicking comments.

I also tried in SC and saw this:
image

So it seems in some cases the text is not fully visible. And if I understand correctly this flag is not to be used on the activity page since we already have the legend but on the new dashboard page that you have right ?

philippe-perrin-sonarsource (approved):
LGTM. Although trying to deal with big leak increase a lot the complexity. Why not displaying the "small leak" design all the time?

components/charts/AdvancedTimeline.tsx
jeremy-davis-sonarsource: It's not necessary for the structure of the svg, so might as well not add one :shrug:

PR #141, opened 2020-09-18
SC-2335 Add not computed QG level

antonio-garcia-sonarsource:
Adding an extra QG Level for not computed as it is important for SC to show the user that it wasn't computed because of missing code period.

https://jira.sonarsource.com/browse/SC-2335

**Checklist**

* [ ] Functional validation
* [ ] Documentation update
* [ ] Update changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


gregaubert (approved):
:+1:

PR #145, opened 2020-10-29
SC-2527 Add DarkHelpTooltip component 

gregaubert:
We are changing a bit the style of our help icon so that they look better on some background colors, we also want them bigger.

**Checklist**

* [x] Functional validation
* [ ] Documentation update
* [x] Update changelog


components/controls/HelpTooltip.tsx
```diff
@@ -20,37 +20,43 @@
...
-            {(theme) => <HelpIcon fill={theme.colors.gray71} size={12} />}
-          </ThemeConsumer>
-        )}
-      </span>
-    </Tooltip>
+export function DarkHelpTooltip({ size = 12, ...props }: Omit<Props, 'children'>) {
+  return (
+    <HelpTooltip {...props}>
+      <HelpIcon fill="rgba(0,0,0,0.25)" fillInner="white" size={size} />
+    </HelpTooltip>
```
jeremy-davis-sonarsource: NP: Why not add a `dark` flag in the props of HelpTooltip?
NP: Add `rgba(0,0,0,0.25)` & `white` to the theme?

jeremy-davis-sonarsource (commented):
I'll do the functional check on SQ

jeremy-davis-sonarsource (approved):
Haven't checked what the dark version looks like, but the normal one seems unchanged :+1:

components/controls/HelpTooltip.tsx
gregaubert: 1. I don't know actually, I didn't think about it xD I wonder if I should do it with a prop instead ? At least like this we can ignore the children prop which would not work with the dark prop... wdyt ?

2. Ok, I wanted but avoided doing this because I thought it would force you on SQ to add variables you don't care about...

jeremy-davis-sonarsource: 1. Ah, I see what you mean. Maybe the problem is the HelpTooltip itself, because if you provide children, it kinda becomes a generic Tooltip :shrug:  
2. We wouldn't need to define them: they would be undefined, but we don't use the new variant anyway, so I think it's fine.

mathieu-suen-sonarsource (approved):
LGTM

components/controls/HelpTooltip.tsx
gregaubert: 1. Ah yeah you are right, there is not much added by the component if you pass a children... But in SC this behavior is used in quite some places... So I'll just leave it like this I think.
2. All good then :) I'll know for next time!

PR #146, opened 2020-11-27
SC-2661 Add light version of button primary

gregaubert:
We need a light version of the primary button so that it works on dark background too

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,12 @@
 
 ## Unreleased
 
+- SC-2661 Add Light version of button primary
+
 ## 1.0.26
 
 - SC-2527 Add DarkHelpTooltip component and allow to customize inner question mark color of HelpIcon
+- SC-2527 Add "white" and "transparentBlack" to the theme
 
 ## 1.0.25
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,9 +2,12 @@
 
 ## Unreleased
 
+- SC-2661 Add Light version of button primary
+
 ## 1.0.26
 
 - SC-2527 Add DarkHelpTooltip component and allow to customize inner question mark color of HelpIcon
+- SC-2527 Add "white" and "transparentBlack" to the theme
```
gregaubert, line 10: Those 2 variables were added in the previous version but not documented in the changelog, it's better if they appear in it since you should know that you have to add them in SQ theme once you update past this version.

components/controls/buttons.css
```diff
@@ -54,11 +54,17 @@
...
-  color: #fff;
+  color: var(--white);
 }
 
 .button-primary:hover {
   background: var(--veryDarkBlue);
+  border-color: var(--veryDarkBlue);
+}
+
+.button-primary.button-light:hover {
```
gregaubert, line 65: So it looks like this with the button light class:
image

And without:
image

jeremy-davis-sonarsource: looks good!

jeremy-davis-sonarsource (approved):
LGTM 
Didn't run it since we have no dark background on SQ.

PR #149, opened 2021-01-08
SONAR-14317 Replacement patterns unescaped in replace method

jeremy-davis-sonarsource:

### DIFF
--- a/helpers/l10n.ts
+++ b/helpers/l10n.ts
@@ -39,7 +39,7 @@ export function translateWithParameters(
   if (message) {
     return parameters
       .map((parameter) => String(parameter))
-      .reduce((acc, parameter, index) => acc.replace(`{${index}}`, parameter), message);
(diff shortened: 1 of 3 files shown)


helpers/l10n.ts
```diff
@@ -39,7 +39,7 @@ export function translateWithParameters(
   if (message) {
     return parameters
       .map((parameter) => String(parameter))
-      .reduce((acc, parameter, index) => acc.replace(`{${index}}`, parameter), message);
+      .reduce((acc, parameter, index) => acc.replace(`{${index}}`, () => parameter), message);
```
gregaubert, line 42: You actually don't need the `` `${parameter}` `` like I do because you have the `.map((parameter) => String(parameter))` just before the reduce, and I dropped it on my PR.

gregaubert (approved):
:+1:

PR #150, opened 2021-02-09
SONAR-12180 Add a ARIA role to all global success and error messages

wouter-admiraal-sonarsource:


mathieu-suen-sonarsource (approved):
LGTM I have not test it with proper screen reader but at least it does not affect normal reader.

PR #151, opened 2021-02-26
Hardening, batch II

wouter-admiraal-sonarsource:
This is required for https://github.com/SonarSource/sonar-enterprise/pull/3815.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,11 @@
 
 ## Unreleased
 
+- SONAR-12056 Indicate required fields for forms 
+
 ## 1.0.31
+
+- SONAR-12180 Add a ARIA role to all global success and error messages
 
 ## 1.0.30
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - SONAR-12056 Indicate required fields for forms 
+- SONAR-11959 Colorblind-friendly coverage treemaps
 
 ## 1.0.31
 
--- a/components/charts/ColorGradientLegend.tsx
+++ b/components/charts/ColorGradientLegend.tsx
@@ -19,27 +19,29 @@
  */
 import { ScaleLinear, ScaleOrdinal } from 'd3-scale';
 import * as React from 'react';
+import { ThemeConsumer } from '../theme';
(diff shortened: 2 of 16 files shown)


CHANGELOG.md
```diff
@@ -2,8 +2,13 @@
 
 ## Unreleased
 
+- SONAR-12056 Indicate required fields for forms 
+- SONAR-11959 Colorblind-friendly coverage treemaps
+
 ## 1.0.31
 
+- SONAR-12180 Add a ARIA role to all global success and error messages
+
```
jeremy-davis-sonarsource, line 11: What happened here?

components/charts/ColorGradientLegend.tsx
```diff
@@ -49,45 +51,81 @@ export default function ColorGradientLegend({
...
+                    x2={i}
+                    y2="30"
+                    style={{ stroke: colors.gray71, strokeWidth: NA_SPACING }}
+                  />
+                  <line
+                    x1={i + NA_SPACING}
+                    y1="0"
+                    x2={i + NA_SPACING}
+                    y2="30"
+                    style={{ stroke: colors.gray60, strokeWidth: NA_SPACING }}
```
jeremy-davis-sonarsource, line 84: NP: Are these values really tied to the ones below? I mean, hey are the same value, but should they be the same constant?

jeremy-davis-sonarsource (approved):
LGTM, WFV

CHANGELOG.md
wouter-admiraal-sonarsource: I forgot to update the Changelog last time 🤷

components/charts/ColorGradientLegend.tsx
wouter-admiraal-sonarsource: You mean, if it makes sense to have `strokeWidth: NA_SPACING` being tied to `i + NA_SPACING`? If so, yes, it allows us to keep an even pattern of 4px wide strokes, spaced 0px apart.

jeremy-davis-sonarsource: I meant tied to the `rect` and `text` below, but yeah, I see the pattern is actually used by the `rect`, so nevermind :+1:

